PR #4, opened 2016-06-09
Add model field translations and documentation for the get versions r…

BernhardPosselt:
…equest


BernhardPosselt:
@adsworth @LukasReschke

adsworth:
I there any specific reason to return all languages and let the client handle it? I can understand why it necessary for the installed apps, but the list of available apps is only seen by the admin and then returning the admins accepted language should be enough. When an app is installed we'd need all available languages, but at that point that information would come from the app itself wouldn't it.

BernhardPosselt:
The idea is to cache the file in Nextcloud based on Http headers at some point. If you only serve one translation, Nextcloud needs to iterate through all supported languages and download the whole spec again which is a lot of duplication. They could also lazy load it but I think it's probably faster to just keep one version.

adsworth:
OK, makes sense. In larger installations I guess there can be admins with different languages. Is the a situation where the information from this request is seen by normal users, I can't see one but then again I'm not a nextcloud expert?

BernhardPosselt:
Normal users won't see this, the API for now is purely for feeding Nextcloud app information for the apps page and for letting devs quickly upload new releases, whose meta information is then parsed and fed into the web interface and API :)

The web interface will use different API routes where we ship only one translation.

BernhardPosselt:
BTW: If you think the pull reuqest is ok, you can hit the merge button :)

Issue #2, opened 2016-06-08
Sign up form using recaptcha


BernhardPosselt:
As a developer I wan't to sign up so I'm able to submit app packages.

The basic sign up process is very simple for now:
* Use a [ModelForm](https://docs.djangoproject.com/en/1.9/topics/forms/modelforms/) to create a form for auth.User
* Explicitely whitelist the fields that should be shown
* Install [django-recaptcha](https://github.com/praekelt/django-recaptcha#field) from [pypi](https://pypi.python.org/pypi/django-recaptcha) in the venv and add it to requirements/production.txt (tip: pip freeze will output all installed packages in the virtual env)
* Add the custom field to the ModelForm and get tokens from Google. Put the tokens into nextcloudappstore/local_settings.py
* Create a simple template with the form and use a class based view ([CreateView](https://docs.djangoproject.com/en/1.9/ref/class-based-views/generic-editing/#django.views.generic.edit.CreateView) ?)
* The form should be served on /register
* You also need to include the js script from google in the header. It should be included only on the register page so maybe go for template inheritance where you add can add extra head tags by inheriting from it.

It could be that the plugin does not work with Django 1.9, in that case look for a compatible solution or just create a widget using the [recaptcha](https://pypi.python.org/pypi/recaptcha-client) Python library

cc @adsworth 

@LukasReschke for input

BernhardPosselt:
Implemented

PR #7, opened 2016-06-09
Add user registration, login and logout

adsworth:
This implements the requirements in #2.


adsworth:
Oh, bootstrap is pulled in via CDN, so it's easy to remove when someone get going on the frontend and doesn't pollute the repository .

adsworth:
the builds failed because there are two new settings:
RECAPTCHA_PUBLIC_KEY and RECAPTCHA_PRIVATE_KEY.

I only added them to my local_settings.py as they are similar to the SECRET_KEY, individual for each installation. How can the Travis build handle that or do we need to add them with no value to settings.py

nextcloudappstore/core/views.py
```diff
@@ -1,3 +1,7 @@
 from django.shortcuts import render
+from django.views.generic import View
 
-# Create your views here.
+
+class HomeView(View):
```
BernhardPosselt: Could probably be even shorter by subclassing TemplateView https://docs.djangoproject.com/en/1.9/ref/class-based-views/base/#templateview

nextcloudappstore/core/user/views.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib import messages
+from django.shortcuts import render
+from django.utils.translation import ugettext as _
+
+import registration.backends.simple.views 
+
+from nextcloudappstore.core.user.forms import RegistrationFormRecaptcha
+
+
+class RegistrationView(registration.backends.simple.views.RegistrationView):
```
BernhardPosselt: I'd probably import that instead of using the full absolute path :)

nextcloudappstore/core/user/views.py
```diff
@@ -0,0 +1,19 @@
...
+    
+    form_class = RegistrationFormRecaptcha
+
+    def get_success_url(self, user):
+        return u'/'
```
BernhardPosselt: Should already be unicode in Python 3 right? Example maybe used Python 2? Am I wrong?

nextcloudappstore/core/user/apps.py
```diff
@@ -0,0 +1,5 @@
+from django.apps import AppConfig
+
+
+class UserConfig(AppConfig):
+    name = 'nextcloudappstore.core.user'
```
BernhardPosselt: we should add a verbose name :)

BernhardPosselt:
Good question, I think we should add them with no value for now, just adjust the make file https://github.com/nextcloud/appstore/blob/master/Makefile#L27

:+1: 

Very good pull request. Translations are missing atm but I think it's just boilerplate text for now and we can add it later :)

nextcloudappstore/core/user/views.py
BernhardPosselt: Ah right, probably even better would be to use 

    return reverse('home')

adsworth: Dang, I wanted todo that but forgot. :)

nextcloudappstore/core/user/views.py
adsworth: Yes, I though about that as well, but...
`from registration.backends.simple.views.RegistrationView import RegistrationView`
doesn't work since I want our view to also be called RegistrationView. I could do 
`from registration.backends.simple.views.RegistrationView import RegistrationView as ....`
but I had no idea what to import it as so I opted for the abosulte path. :)

Do you have a suggestion?

BernhardPosselt: Ah right, I'll leave it up to you :)

adsworth:
I presume string should all be in english in the source. Correct?

BernhardPosselt:
Yes thinking more about html :)

BernhardPosselt:
:+1: looks good to me

Once the merge conflicts are resolved you can hit the merge button

BernhardPosselt:
Great job, tested it and works fine :)

Bootstrap styles also look good

PR #8, opened 2016-06-10
Fix #5, Fix#6

### LINKED ISSUE #5: App/AppRelease deletion authorizations
Currently we've got an app owner and app collaborators. Both are allowed to publish and delete specific apps and their releases. Are we fine with this? If so we can close this ticket :)

@LukasReschke @adsworth @jancborchardt

BernhardPosselt:
@adsworth @LukasReschke


docs/api.md
```diff
@@ -101,3 +102,44 @@ This is the route that will return all releases to display inside Nextcloud's ap
...
+* **Authentication**: Basic
+
+* **Returns**:
+  * **HTTP 204**: If the app was deleted successfully
+  * **HTTP 400**: If the app could not be found
+  * **HTTP 401**: If the user is not authenticated
+  * **HTTP 403**: If the user is not authorized to delete the app
+
+## Delete An App Release
+Only an app owner or co-maintainers are allowed to delete an app release. The owner is set to the first user that pushes a release to the store, co-maintainers can be added.
```
adsworth: This should read "Only app owners or co maintainers are allowed..."

and how about "Owner is the user that pushes the first release of an app to the store,..."

docs/api.md
```diff
@@ -101,3 +102,44 @@ This is the route that will return all releases to display inside Nextcloud's ap
...
+
+
+* **Url**: DELETE /api/v1/apps/{app-id}/{app-version}
+  * **Parameters**:
+    * **app-id**: app's id, lower case ASCII characters and underscores are allowed
```
adsworth: I'm pretty sure apps is correct in this case.

BernhardPosselt: yes, app id would work

docs/api.md
BernhardPosselt: Right, will adjust it.

nextcloudappstore/core/models.py
```diff
@@ -47,8 +47,11 @@ class Meta:
     def __str__(self):
         return self.name
 
-    def allows_editing(self, user):
```
adsworth: Django permissons uses can_*. allows_* suggest that the user allows someone to edit. It doesn't sound like a check if edting is allowed.

BernhardPosselt: good point, can_edit and can_delete sounds much better

BernhardPosselt:
Issues are fixed, however github does not show the newest diff which is weird.

PR #21, opened 2016-06-12
Restructuring

BernhardPosselt:
@adsworth @LukasReschke 

TL;DR: 
* move to json fixtures because of bugs in the yaml lib
* add some utility functions in facades.py which im using atm
* rename production.txt to base.txt
* update app upload docs
* add licenses, recommendations and checksums to the data model
* add ordering parameter to screenshots to allow users to control which screenshot will be presented first 
* add pytz because its recommended
* add requests because there is no straight forward non low level way to stream downloads and Python http sucks


docs/api.md
```diff
@@ -103,6 +111,8 @@ Translated fields are stored inside a translations object. They can have any siz
 
 Required versions (minimum and maximum versions) are stored in String fields. If a field is empty, this means that there is no version requirement. Minimum and maximum versions are inclusive, e.g. if your app works only on PHP 5.5 and 5.6, the minimum version will be 5.5.0 while the maximum version will be 5.6.*
 
+The checksum is generate using the archive via sha256sum
```
adsworth: ...generated...

docs/api.md
```diff
@@ -143,3 +152,133 @@ Only app owners or co-maintainers are allowed to delete an app release. The owne
...
+        curl -X PUT -u "user:password" http://localhost:8000/api/v1/apps/news/releases -H "Content-Type: application/json" -d '{"download":"https://example.com/release.tar.gz"}'
+
+* **Returns**:
+  * **HTTP 200**: If the app release was update successfully
+  * **HTTP 201**: If the app release was created successfully
+  * **HTTP 400**: If the app contains invalid data, is too large or could not be downloaded
+  * **HTTP 401**: If the user is not authenticated
+  * **HTTP 403**: If the user is not authorized to delete the app release
+
+If there is not yet an app with the given app id yet, a new app is created and the owner is set in to the logged in user. Then the **info.xml** file which lies in the compressed archive's folder **app-id/appinfo/info.xml** is being parsed and validated.
```
adsworth: you have "yet" in there twice. Maybe reword to "If no app with the given app id exists,..."

docs/api.md
```diff
@@ -143,3 +152,133 @@ Only app owners or co-maintainers are allowed to delete an app release. The owne
...
+
+Afterwards the app and app release will be created or updated, depending on if they already exist. Both owners and co-maintainers are allowed to upload new releases.
```
adsworth: This is different to what you write above. Above you wirte "if not exist create app and download archiv and read info.xml. Afterwards app and release will be created. 
So what actually happen download and then create/update or create and then download?

PR #24, opened 2016-06-12
app list view and app detail added

adsworth:


nextcloudappstore/core/templates/home.html
```diff
@@ -33,7 +24,7 @@
 {% for app in recommended_apps %}
   <div class="col-sm-4">
     <div class="well">
-      <h3><a href="http://getbootstrap.com">{{ app.name }}</a></h3>
+      <h3><a href="{%url 'app_detail' id=app.id %}">{{ app.name }}</a></h3>
```
BernhardPosselt: do we go for _ or - for separating words?

nextcloudappstore/core/views.py
```diff
@@ -1,7 +1,36 @@
...
+
+class HomeView(CategoryContextMixin,
+               RecommendedAppsContextMixin, TemplateView):
     template_name = "home.html"
+
+
+class AppDetailView(CategoryContextMixin, DetailView):
+    model = App
+    template_name = 'app/detail.html'
+    slug_field = 'id'
```
BernhardPosselt: I think you can name the parameter in the url slug and it just works :)

BernhardPosselt:
Well, whatever looks good :)

As for separating words in URLs, would you go for - or _?

nextcloudappstore/core/views.py
adsworth: No, it doesn't. I tried that and when getting DetailView complains that it needs either a pk or slug kwarg.

PR #26, opened 2016-06-14
email confirmation and social login github and bitbucket

### LINKED ISSUE #3: Sign up should be confirmed via E-Mail
As a user I want to confirm my sign up by clicking on a link inside an E-Mail so I can verify that I've used the correct E-Mail for future password resets.

Basic implementation should start where #2 left off and use http://django-registration.readthedocs.io/en/2.1/quickstart.html or a similarily well maintained and popular extension to implement the registration flow. 

The used library should be installed via pip and added to requirements/production.txt

adsworth:
this fixes #3


adsworth:
@BernhardPosselt @LukasReschke ready for review.

nextcloudappstore/core/templates/account/signup.html
```diff
@@ -0,0 +1,42 @@
...
+{% load socialaccount %}
+
+{% load i18n %}
+
+{% block head_title %}{% trans "Signup" %}{% endblock %}
+
+{% block content %}
+<h1>{% trans "Sign Up" %}</h1>
+
+<p>{% blocktrans %}Already have an account? Then please <a href="{{ login_url }}">log in</a>.{% endblocktrans %}</p>
```
BernhardPosselt: Do we have a custom url tag for the login url? like {% url 'django.contrib.auth.views.login' %} ?

nextcloudappstore/core/templates/account/signup.html
```diff
@@ -0,0 +1,42 @@
...
+</div>
+{% endif %}
+
+<form class="signup" id="signup_form" method="post" action="{% url 'account_signup' %}">
+  {% csrf_token %}
+  <table>
+  {{ form }}
+  </table>
+  {% if redirect_field_value %}
+  <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}" />
```
BernhardPosselt: Where do we get this value from? If it is a GET paramter, a user could be sent a form with an evil redirect site

nextcloudappstore/core/user/forms.py
```diff
@@ -1,11 +1,11 @@
...
 
-class RegistrationFormRecaptcha(RegistrationFormUniqueEmail):
+class SignupFormRecaptcha(forms.Form):
     """integrate a recaptcha field."""
     recaptcha = ReCaptchaField()
 
-    def clean(self):
-        self._validate_unique = True
-        super().clean()
+    def signup(self, request, user):
```
BernhardPosselt: What's the reason that this method returns None and why do we need it ?

BernhardPosselt:
Wow, looks pretty easy with allauth.

BernhardPosselt:
Branch has conflicts btw

nextcloudappstore/core/templates/account/signup.html
adsworth: Yes, there is account_login which is supplied by the allauth.account.urls.
The variable login_url is supplied by the login view allauth.account.views.LoginView and generate using reverse.

nextcloudappstore/core/templates/account/signup.html
adsworth: Yes that could possibly be the case. The value comes directly from request.POST or request.GET.

It is set in https://github.com/pennersr/django-allauth/blob/master/allauth/account/views.py#L207 and get_request_param is defined in https://github.com/pennersr/django-allauth/blob/master/allauth/utils.py#L272.

I'll check django.contrib.auth.views.login to see what they do there. I think remember that that view was hardend against malicous redirects.

nextcloudappstore/core/user/forms.py
adsworth: We need it becuase allauth assume that if you create your own signup form, then you need an extra hook to do what needs doing when you sign up a user. I think the social signup view uses this method to acutally create the django user and hook the social token to the user. If the method isn't defined allauth raises an exception.

I guess I could make a no-op out of it and just you pass instead of return None.

adsworth:
grr,confilcts.

nextcloudappstore/core/templates/account/signup.html
adsworth: OK, [django.contrib.auth.views.login](https://github.com/django/django/blob/master/django/contrib/auth/views.py#L53) validates the url using [django.utils.http.is_save_url](https://github.com/django/django/blob/master/django/utils/http.py#L280).

Hmm, fork, patch, create pull request for allauth and then install from the fork for now. Or what should we do?

BernhardPosselt: It's ok if it points to a location on our server I think @LukasReschke

adsworth: I can't be posted by the user to any url. So we should harden it.

adsworth: altouhg request.POST.get and request.GET.get escape any HTML characters. So a user could force a redirect to another domain but I think injecting malicious JS is not possible.

BernhardPosselt: A misunderstood then :)

adsworth: Some diggin later. allauth does check where the redirect points to just before the actual redirect happens.

All requests to redirect run through this piece of code: https://github.com/pennersr/django-allauth/blob/master/allauth/account/utils.py#L46 which in end effect calls the Django function is_safe_url.

PR #27, opened 2016-06-15
Description for Category model

### LINKED ISSUE #22: add description field to category model
I think a description of the category would be a nice addition.

Thoughts?

@BernhardPosselt @LukasReschke

clone1612:
Attempt to add a description field to the Category model in order to fix #22. A description field for categories was already present in the html file so little had to be changed.

I'm further familiarizing myself with the codebase, so if I missed any code conventions feel free to tell.


BernhardPosselt:
Thanks! Looks good but please squash the migrations :)

Until we've got something stable it makes no sense to pile up migrations, especially because its slow ;)

clone1612:
That makes a lot of sense! Should be fixed now.

BernhardPosselt:
👍

PR #30, opened 2016-06-17
Apps etags

### LINKED ISSUE #12: Add Http Caching via Etag to the api:apps route
Both apps and app releases provide a lastModified field. This field contains the modified date down to miliseconds.

If a request comes in, the http headers should be checked for **If-None-Match**. A lookup in the database should determine if there is an app or release with a newer last modified date. If there is none, a Http 304 should be returned without further querying the database.

If there is a newer entry, the entire thing should be served again. An Etag header should be set on the response which contains the latest lastModified from both apps and releases.

BernhardPosselt:
Ready to review

@LukasReschke @adsworth

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -3,7 +3,7 @@ This is a draft and will continuously be updated :)
 
 
 ## Get All Apps And Their Releases For a Platform Release
-This is the route that will return all releases to display inside Nextcloud's apps admin area. Future releases will add Http caching.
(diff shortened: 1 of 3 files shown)


adsworth:
Looks good.

BernhardPosselt:
You closed it instead of merged it ;D

adsworth:
Ups, sorry.

PR #33, opened 2016-06-18
Add cors for api

### LINKED ISSUE #13: API routes should be enabled for CORS
Only API routes should be whitelisted for CORS

We need to make sure that CORS and session auth are mutually exclusive

The solution is probably to integrate https://github.com/ottoyiu/django-cors-headers/

BernhardPosselt:
Fix #13 

Reade to review @adsworth @LukasReschke


adsworth:
Looks good.

PR #37, opened 2016-06-19
order app releases by version descending per default

adsworth:


nextcloudappstore/core/models.py
```diff
@@ -106,6 +106,7 @@ class Meta:
         verbose_name = _('App Release')
         verbose_name_plural = _('App Releases')
         unique_together = (('app', 'version'),)
+        ordering = ['-version']
```
BernhardPosselt: Does this work with semantic releases? AFAIK it uses natural ordering so 13.0.0 is considered lower than 2.0.0

adsworth: as it is a charfield that should work since 1 is smaller than 2

BernhardPosselt: AFAIK the ordering field is translated to Sql. ORDER BY for ints is fine, however for chars I think it works on comparing char for char. 

Lets go for 13.0.0 and 2.0.0: First it compares 1 and 2. Since 1 is lower than 2, the string comparison will return that the 13.0.0 version is lower than 2.0.0.

Can you try it out :)

Also check https://blog.codinghorror.com/sorting-for-humans-natural-sort-order/

adsworth: I misunderstood you. I thought you meant semantic versioning also uses natural ordering.
If it doesn't then this won't work.

I just saw that semantic_version has django fields. Maybe we should use them, then we can at least use sort() to get the correct order.

It even has a SpecField which could be used to replace the min_* and max_*  fields. https://pypi.python.org/pypi/semantic_version/

BernhardPosselt: Oh, that'd be great. We'd need to migrate incoming versions to a spec but that would make sense anyways. If we wanted to allow -beta suffixes we'd also just have to change the xsd which would be awesome :)

BernhardPosselt: I'd take care of the import transforms if that's ok for you. Can you adjust the models?

adsworth: Yeah, I can do the models.

so :
-AppRelease.version becomes semantic_version.django_fields.VersionField 
- php_min_version and php_max_version becomes php_version_spec = semantic_version.django_fields.VersionField 
- dito with the other (min|max)_version fields i.e. platform, database and php extension.

nextcloudappstore/core/api/v1/tests/test_release_importer.py
```diff
@@ -6,6 +6,7 @@
 from nextcloudappstore.core.facades import read_relative_file
 from nextcloudappstore.core.models import App, Screenshot, Database
 from pymple import Container
+from semantic_version import Version, Spec
```
adsworth: this could be dropped, but then again it doesn't hurt.

adsworth:
looks good to me.

BernhardPosselt:
Thanks :)

BTW unused imports cause bugs later on,  good call ;)

PR #68, opened 2016-06-27
development setup tweaks.

adsworth:
- don't rebuild inital migrations on every resetup
- ignore settings.development in git, everybody has their own setting so have that in git is just annoying
- add a template settings.development module and docs telling the devloper to review that module and adjust according to their needs.
- new make target initmigrations to recreate all inital migrations.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,6 @@ nextcloudappstore/local_settings.py
 *.sqlite3
 *.wp[ru]
 nextcloudappstore/settings/production.py
+nextcloudappstore/settings/development.py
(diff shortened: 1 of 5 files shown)


adsworth:
@BernhardPosselt I have an issue with the Makefile.
On my setup I need to use echo -e for echo to interpret backslash escapes, but then travis fails.
Also having the default settings/development.py all in one echo line feels kind of clunky, but I didn't find another solution. Using a multiline \ echo statement always caused an extra space at the beginning of the lines, which gives a python syntax error.

PR #70, opened 2016-06-29
initial token auth support

adsworth:
- token auth support in the API views.
- create a when a local user signs up.


BernhardPosselt:
Looks good but for all the auth things it would be awesome if we also had tests.

Simply create a _login_token() method here https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/tests/api.py#L16 much like the _login method (unless you find an easier way).

Then copy this test method and swap out self._login() with self._token_login()

Then do the same here: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/tests/test_app_release.py#L28 and here: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/tests/test_app_release.py#L87

Auth tests have saved me a couple of times so I think its worth it :)

Issue #65, opened 2016-06-27
Nightly support


MorrisJobke:
- nightlies don't have a separate version number but a flag:

```
curl -X POST -u "user:password" http://localhost:8000/api/v1/apps/releases -H "Content-Type: application/json" -d '{"download":"https://example.com/release.tar.gz", "nightly":true }'
```

- this is also listed in the "get all apps" API with a `nightly: true` attribute https://nextcloudappstore.readthedocs.io/en/latest/restapi.html#get-all-apps-and-releases
- upload of a new nightly will delete the previous one for that app
- this allows to upgrade to a nightly (needs to be invoked by the admin and can be undone -> next regular release of the app will be installed)

BernhardPosselt:
General question: We internally store a nightly as x.x.x-nightly since we need a unique version number and app id. Deleting a nightly also works like this: 

    DELETE http://localhost:8000/api/v1/apps/news/releases/9.0.0-nightly

This needs to be done because we'd otherwise have to do weird stuff like split the version:

    DELETE http://localhost:8000/api/v1/apps/news/releases/9.0.0/nigthly


However in our REST API we split up the version number and the nightly modifier into a version and flag:

```json
{
    "version": "9.0.0",
    "nightly": true
}
```

@MorrisJobke @LukasReschke what do you think in terms of consistency and API usability? Do prefer the nightly flag or would you opt for a combined version like the following, which is perfectly fine regarding semver?

```json
{
    "version": "9.0.0-nightly"
}
```

MorrisJobke:
> @MorrisJobke @LukasReschke what do you think in terms of consistency and API usability? Do prefer the nightly flag or would you opt for a combined version like the following, which is perfectly fine regarding semver?

I would go for the nigthly flag, because otherwise the version in the API and the version in the info.xml are different.

janibonnevier:
I think a flag makes it clearer that a nightly release will result in some special behaviour, namely that it replaces other nightlies, as opposed to just being a normal release with "-nightly" appended to it.

PR #71, opened 2016-06-29
Add support for nightlies

janibonnevier:
#65 There might be a bug here that needs to be fixed before merging.


BernhardPosselt:
LGTM :) :+1:

PR #73, opened 2016-06-29
Fix  #65

### LINKED ISSUE #65: Nightly support

BernhardPosselt:
@janibonnevier @adsworth @LukasReschke


adsworth:
looks good to me.

PR #74, opened 2016-06-30
importer cleanup, url, view and template cleanup, do not delete app r…

BernhardPosselt:
…elations if they are not imported


nextcloudappstore/core/templates/app/base.html
```diff
@@ -13,7 +13,7 @@
             </li>
             {% for cat in categories %}
             <li class="{% if cat.id == current_category.id %}active{% endif %}">
-                <a href="{% url 'category_list_category' cat.id %}">{{cat.name }}</a></li>
+                <a href="{% url 'app-list' cat.id %}">{{ cat.name }}</a></li>
```
adsworth: I find app-list really confusing here. I would at least use app-list-by-category. I'd guess we'll at least get an app-list-by-developer route at some point.

BernhardPosselt: :+1: right

nextcloudappstore/urls.py
```diff
@@ -1,14 +1,16 @@
...
-    url(r'^$', AppListView.as_view(), {'id': None}, name='home'),
+    url(r'^$', CategoryAppListView.as_view(), {'id': None}, name='home'),
     url(r'^', include('allauth.urls')),
-    url(r'^categories/(?P<id>[\w]*)$', AppListView.as_view(),
-        name='app-list'),
-    url(r'^app/(?P<id>[\w]+)$', AppDetailView.as_view(), name='app-detail'),
+    url(r'^categories/(?P<id>[\w]*)$', CategoryAppListView.as_view(),
+        name='category-app-list'),
+    url(r'^app/(?P<id>[\w_]+)$', CategoryAppDetailView.as_view(),
+        name='category-app-detail'),
```
adsworth: shouldn't this just be app-detail?

BernhardPosselt: right!

PR #75, opened 2016-06-30
Add token auth docs

BernhardPosselt:


docs/restapi.rst
```diff
@@ -5,6 +5,19 @@ A REST API for publishing and deleting app releases has been built into the stor
...
+
+Several routes require authentication. The following authentication methods are supported:
+
+* **Basic**: Http header where **CREDENTIALS** is ``base64encode('user:password')``::
+
+    Authorization: Basic CREDENTIALS
+
+* **Basic**: Http header where **TOKEN** is a token which can be looked up in your profile::
+
+    Authorization: Bearer TOKEN
```
adsworth: It's Authorization: Token TOKEN, though that can be changed to Bearer. We would need to subclass TokenAuthentication and set the keyword class attribute. See also http://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication

docs/restapi.rst
```diff
@@ -5,6 +5,19 @@ A REST API for publishing and deleting app releases has been built into the stor
...
+Authentication
+--------------
+
+Several routes require authentication. The following authentication methods are supported:
+
+* **Basic**: Http header where **CREDENTIALS** is ``base64encode('user:password')``::
+
+    Authorization: Basic CREDENTIALS
+
+* **Basic**: Http header where **TOKEN** is a token which can be looked up in your profile::
```
janibonnevier: I believe this line should say **Token** instead of **Basic**.

docs/restapi.rst
BernhardPosselt: Remembered it wrongly, its token now :)

docs/restapi.rst
BernhardPosselt: damn xD

PR #76, opened 2016-07-01
Add CSP

BernhardPosselt:
Produces the following HTTP header in dev:

    content-security-policy:"img-src *; default-src 'none'; script-src 'self'; connect-src 'self'; font-src 'self'; style-src 'self'"

and when following the docs the following in production:

    content-security-policy:"img-src https; default-src 'none'; script-src 'self'; connect-src 'self'; font-src 'self'; style-src 'self'"


LukasReschke:
🚀 👍

PR #90, opened 2016-07-02
Remove recommendations since ratings will not be part of the first st…

BernhardPosselt:
…able release

@janibonnevier @LukasReschke @adsworth @oparoz


BernhardPosselt:
This will prevent us from breaking the api once we settled on a rating system

Issue #82, opened 2016-07-01
App search


BernhardPosselt:
As a user you should be able to search apps. The search field should be rendered in the navigation or banner (inline it) and should search app's titles, descriptions and comments. The resulting list should look and work the same as the app list (meaning you can also sort based on various criteria).

The first implementation should be very simple.

adsworth:
Please don't search in the comments. I think that would cause to many false positives. If I search for bookmark then I don't want apps to appear were a user commented "I saved a bookmark" or something similiar. If you want comments searchable then make it optional with a checkbox.
Personally I think I wold only ever want to search the comments of one specific app. I don't see the value of searching the comments of all apps at once.

BernhardPosselt:
Fine by me :) This can be pushed back in case we need some advanced search

PR #93, opened 2016-07-04
Simple app search

janibonnevier:
#82 

I have not done much to make sure that this is secure, as I'm unsure about how much Django does automatically.

Includes some changes to the banner layout css.

### DIFF
--- a/nextcloudappstore/core/templates/nav.html
+++ b/nextcloudappstore/core/templates/nav.html
@@ -3,6 +3,9 @@
 <div class="banner">
     <div class="container">
         <a class="brand" href="/"><img class="logo" src="{% static 'assets/img/logo_nextcloud_white.svg' %}" title="nextcloud.com" itemprop="logo"></a>
+        <form method="get" action="" class="search-form">
+            <input type="text" placeholder="Search" class="form-control search-box">
+        </form>
     </div>
 </div>
 <nav class="navbar navbar-default navbar-static-top">
--- a/nextcloudappstore/core/templates/nav.html
+++ b/nextcloudappstore/core/templates/nav.html
@@ -3,8 +3,13 @@
 <div class="banner">
     <div class="container">
         <a class="brand" href="/"><img class="logo" src="{% static 'assets/img/logo_nextcloud_white.svg' %}" title="nextcloud.com" itemprop="logo"></a>
-        <form method="get" action="" class="search-form">
-            <input type="text" placeholder="Search" class="form-control search-box">
+        <form method="get" action="{{ request.path }}" class="search-form">
+            {% if current_category %}
+            <input type="text" name="search" placeholder="Search in {{ current_category }}" class="form-control search-box">
+            {% else %}
+            <input type="text" name="search" placeholder="Search all apps" class="form-control search-box">
+            {% endif %}
+            <input type="submit" value="Search" class="search-button">
         </form>
     </div>
 </div>
--- a/nextcloudappstore/core/views.py
+++ b/nextcloudappstore/core/views.py
@@ -1,6 +1,8 @@
 from django.views.generic.detail import DetailView
 from django.views.generic.list import ListView
 from nextcloudappstore.core.models import App, Category
+from django.http import Http404
+from django.db.models import Q
 
 
 class AppDetailView(DetailView):
@@ -19,13 +21,31 @@ class CategoryAppListView(ListView):
     model = App
     template_name = 'app/list.html'
 
-    def get_queryset(self):
+    def get_queryset(self, request):
         category_id = self.kwargs['id']
         queryset = super().get_queryset()
+
         if category_id:
-            return queryset.filter(categories__id=category_id)
-        else:
-            return queryset
+            queryset = queryset.filter(categories__id=category_id)
+
+        if ('search' in request.GET) and request.GET['search'].strip():
+            search_words = request.GET['search'].strip().split()
+            query = None
+
+            for word in search_words:
+                q = Q(translations__name__contains=word) | \
+                    Q(translations__description__contains=word)
+                if query is None:
+                    query = q
+                else:
+                    query = query | q
+
+            queryset = queryset.filter(query)
+
+            # Remove duplicates that for some reason sometimes occur
+            queryset = list(set(queryset))
+
+        return queryset
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
@@ -34,3 +54,23 @@ class CategoryAppListView(ListView):
         if category_id:
             context['current_category'] = Category.objects.get(id=category_id)
         return context
+
+    def get(self, request, *args, **kwargs):
+        self.object_list = self.get_queryset(request)
+        allow_empty = self.get_allow_empty()
+
+        if not allow_empty:
+            # When pagination is enabled and object_list is a queryset,
+            # it's better to do a cheap query than to load the unpaginated
+            # queryset in memory.
+            if (self.get_paginate_by(self.object_list) is not None
+                    and hasattr(self.object_list, 'exists')):
+                is_empty = not self.object_list.exists()
+            else:
+                is_empty = len(self.object_list) == 0
+            if is_empty:
+                raise Http404(
+                    _("Empty list and '%(class_name)s.allow_empty' is False.")
+                    % {'class_name': self.__class__.__name__})
+        context = self.get_context_data()
+        return self.render_to_response(context)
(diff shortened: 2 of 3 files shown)


janibonnevier:
I realized the `action` attribute of the form needs to be set to something else if there is no current category.

nextcloudappstore/core/templates/nav.html
```diff
@@ -3,6 +3,15 @@
 <div class="banner">
     <div class="container">
         <a class="brand" href="/"><img class="logo" src="{% static 'assets/img/logo_nextcloud_white.svg' %}" title="nextcloud.com" itemprop="logo"></a>
+        {% if current_category %}
+        <form method="get" action="{{ request.path }}" class="search-form">
+            <input type="text" name="search" placeholder="Search in {{ current_category }}" class="form-control search-box">
```
adsworth: redisplay the search term after submission.

nextcloudappstore/core/views.py
```diff
@@ -19,13 +21,31 @@ class CategoryAppListView(ListView):
...
+
+            queryset = queryset.filter(query)
+
+            # Remove duplicates that for some reason sometimes occur
+            queryset = list(set(queryset))
+
+        return queryset
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
```
adsworth: add the search terms to the context data. Then you can display an error message in the template if there are search terms available and the object list is empty.

nextcloudappstore/core/views.py
```diff
@@ -34,3 +54,23 @@ def get_context_data(self, **kwargs):
         if category_id:
             context['current_category'] = Category.objects.get(id=category_id)
         return context
+
+    def get(self, request, *args, **kwargs):
```
adsworth: set allow_empty = True in the list view and then get rid of this redefinition.

nextcloudappstore/core/views.py
```diff
@@ -19,13 +21,31 @@ class CategoryAppListView(ListView):
     model = App
     template_name = 'app/list.html'
 
-    def get_queryset(self):
+    def get_queryset(self, request):
```
adsworth: the request is available as self.request here. No need to redefine get and pass it to the method.

janibonnevier:
I'll see if I can do something about escaping "<" and ">".

nextcloudappstore/core/views.py
```diff
@@ -18,19 +20,48 @@ def get_context_data(self, **kwargs):
...
 
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['categories'] = Category.objects.all()
         category_id = self.kwargs['id']
         if category_id:
             context['current_category'] = Category.objects.get(id=category_id)
+        if self.search_terms:
+            context['search_query'] = ' '.join(self.search_terms)
+            context['search_terms'] = self.search_terms
```
janibonnevier: The `search_terms` context data is not used anywhere. Do we keep it in case we need a list of search terms in a template, or do we remove it?

adsworth: then I'd drop it.

nextcloudappstore/core/templates/app/list.html
```diff
@@ -21,6 +22,12 @@
     {% if forloop.counter|divisibleby:2 %}
         <div class="row"></div>
     {% endif %}
+{% empty %}
+    {% if search_query and current_category %}
+        <p class="message no-apps-found">No apps found for search "<em>{{ search_query }}</em>" in {{ current_category.name }}.</p>
```
BernhardPosselt: This needs a blocktrans for translations :)

nextcloudappstore/core/templates/nav.html
```diff
@@ -3,6 +3,20 @@
 <div class="banner">
     <div class="container">
         <a class="brand" href="/"><img class="logo" src="{% static 'assets/img/logo_nextcloud_white.svg' %}" title="nextcloud.com" itemprop="logo"></a>
+        <form method="get"
+              action="{% if current_category %}{{ request.path }}
```
BernhardPosselt: Wouldnt {{ request.path }} alone work too?

nextcloudappstore/core/templates/nav.html
```diff
@@ -3,6 +3,20 @@
...
+                      {% else %}{% url 'home' %}
+                      {% endif %}"
+              class="search-form">
+            <input type="text"
+                   name="search"
+                   value="{{ search_query }}"
+                   placeholder="{% if current_category %}Search in {{ current_category.name }}
```
BernhardPosselt: I'd probably drop the in: {% if current_category %}Search {{ current_category.name }}

Also needs blocktrans

nextcloudappstore/core/templates/nav.html
```diff
@@ -3,6 +3,20 @@
...
+                                {% else %}Search all apps
+                                {% endif %}"
+                   class="form-control search-box">
+            <input type="submit" value="Search" class="search-button">
```
BernhardPosselt: Needs trans

nextcloudappstore/core/templates/nav.html
janibonnevier: That would break the search box on app detail pages.

nextcloudappstore/core/views.py
```diff
@@ -18,19 +20,48 @@ def get_context_data(self, **kwargs):
...
         if category_id:
-            return queryset.filter(categories__id=category_id)
-        else:
-            return queryset
+            queryset = queryset.filter(categories__id=category_id)
+
+        if self.search_terms:
+            query = None
+
+            for term in self.search_terms:
```
BernhardPosselt: Ok, I think I understand now what this section means, you are doing map reduce the hard way :D

Another thing: 
* you probably want to use icontains to not depend on lower and uppercasing
* you probably want to join the search terms using & and not |, otherwise you will get more results if you add more terms when you want the opposite

Assuming that search_terms is a list of strings I'd actually write it that way (you can also get rid of the if self.search_terms):

```python
from functools import reduce
from django.db.models.functions import Lower

def create_search_query(terms):
    predicates = map(lambda t: (Q(translations__name__icontains=t) |
                                Q(translations__description__icontains=t)),
                     terms)
    return reduce(lambda x, y: x & y, predicates, Q())
```

BernhardPosselt: Another thing: do we know if this works with translations?

nextcloudappstore/core/templates/nav.html
BernhardPosselt: Ah right

nextcloudappstore/core/views.py
janibonnevier: I opted for | because of its more forgiving behaviour (you can enter some bad search terms and still find something), but I do see your point.
I have no idea how this works with translations.

BernhardPosselt: Yeah, you usually dont want to search bad terms :D If you add more than one term you normally want to narrow down your search.

BernhardPosselt: As for the translations: maybe this is a good point in time to extract the search logic into a separate function and create some tests for it.

Check out https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/tests/test_versioning.py#L6 and https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/tests/__init__.py

You can create some apps first and set the appropriate language on them. Then run a test search and check if the correct number of apps is being returned :)

janibonnevier: I'll push the changes first and then take a look at it.

janibonnevier:
I figured out why the duplicates sometimes occur. It turns out that all searches are conducted in all translations, and if all search terms appear in multiple translations for an app, the app will appear multiple times in the search results.

adsworth:
I guess then you need to add the current language to the query. So it only searches the active language.
I think this is mentioned in the special notes in the parler readme.
https://github.com/django-parler/django-parler#special-notes

BernhardPosselt:
@janibonnevier hi jani, can we meet on irc :D?

nextcloudappstore/core/models.py
```diff
@@ -56,18 +56,12 @@ def can_delete(self, user: User) -> bool:
 
     @staticmethod
     def search(lang, terms):
```
adsworth: I think the Django way would be to create a manager class and add that method to the manager.

BernhardPosselt: Right, the question really is how that would work out with django-parler, since that one also overrides the default manager afaik and what the benefits would be since its basically only a single query.

BernhardPosselt: TL;DR: do we get anything valuable in return that justifies the manager and does it play well with parler

PR #97, opened 2016-07-05
GUI fixes and tweaks

janibonnevier:
- Center search box vertically
- Fix glitch on narrow screens where there is 5px of nothingness on the right hand side of the site
- Set min-width of site to 230px to prevent glitches
- CSS code styling
- Adjust media query max-width


BernhardPosselt:
:+1:

PR #98, opened 2016-07-05
App list

### LINKED ISSUE #81: App list page
Apps are rendered in a minimal list atm. 

Each app section should display the first screenshot and maybe render the description (in an ellipsis) and the featured flag.

The list should be able to be sorted ascending and descending based on: 
* last_modified date (can we also include app release modified dates?)
* featured flag
* alphabetic name
* rating (if implemented)

The list should not be paginated. ATM the old app store contains a total of 250 apps which is no performance problem. Pagintation just makes the experience worse.

BernhardPosselt:
Fix #81 by adding sorting in the list overview

This PR also adds an additional translated field called summary. The summary is a short description which will be displayed on the app list page. Contrary to the description it will not contain markdown. If the summary field does not exist, the description field is taken.

@adsworth @janibonnevier @LukasReschke


janibonnevier:
I'm getting duplicates in the alphabetical list.

BernhardPosselt:
Hm so order_by also requires as distinct?

janibonnevier:
There seems to be as many duplicates as there are translations.

BernhardPosselt:
And  what I don't get is that the search call already runs a distinct on it

janibonnevier:
Also, non-featured apps appear in the featured list.

janibonnevier:
Adding another `distinct()` on the `order_by()` does not get rid of the duplicates.

UI-wise, I think there should be visual hint suggesting that you can reverse the ordering by clicking "Alphabetical" again, perhaps a small arrow pointing in the current ordering direction. I just discovered it by accident.

BernhardPosselt:
Yeah we need to think about the hint :)

As for featured: do we want sorting or filtering? Currently it's sorting, that's why all apps appear in there

janibonnevier:
In my mind a "Featured" button is definitely a filter, because I don't expect to see non-featured things if I click on it. I do see that ordering is more consistent here considering that the other buttons do ordering, but at the very least it should say which apps are featured and which are not. However, I think a better solution would be to have checkboxes for filters, and have Featured be one of them.

adsworth:
I agree with Jani here. Featured is a filter.

BernhardPosselt:
Check box is a good idea, probably better to remove the sorting then

janibonnevier:
Have not yet fixed the duplicates.

janibonnevier:
Any opinions on arrow directions?

adsworth:
IMHO, triangle down is descending and triangle up is ascending.

janibonnevier:
Thunar at leat does the opposite, and I guess the idea is that the arrow points toward the ascending direction in the list.

adsworth:
you're right, the GNOME file manager does the same.

I'm not particularly biased one way or the other. So do what you think is "right".

Issue #80, opened 2016-07-01
App detail page 


BernhardPosselt:
The app detail page should list the app author, name and description (rendered in markdown) in the appropriate locale, the screenshots, releases and links.

The screenshots should get the most prominent spot and should be shown in a slideshow (non jquery, slim solution preferred. If you don't need JS but only solve it using CSS, even better!)

Vertical layout should be:
* Title
* Gallery
* Description
* Releases page links, one for each platform (Nextcloud 9.0, Nextcloud 9.1) that contains at least one of the app's releases

The authors information should NOT show his email address and should give him just enough space to feel proud but should not steal the focus from the screenshots.

We should also use an icon or a special style to render the featured attribute.

All links like documentation, website, bug tracker etc, should be placed discretely. The old app store does this pretty well, check out https://apps.owncloud.com/content/show.php/Gallery%2B?content=169116

janibonnevier:
Is the idea here to completely ditch the current layout, or just tweak it?

BernhardPosselt:
Ditch mostly.

janibonnevier:
I'm not sure what to do about the Releases tab. Is #79 referring to this tab, or an actual new separate page?

BernhardPosselt:
Releases are a separate ticket.

We don't want a releases tab, we want a list of nextcloud versions that have at least one app release. If you click on one release we should render a list of all releases running on that version where the newest release is on top.

janibonnevier:
And that list should be rendered right there on the app detail page?

BernhardPosselt:
The nextcloud version list, yes, the app version list, no

BernhardPosselt:
I mean we could add two buttons like: download the latest release for this nextcloud version and see all releases for that nextcloud version

BernhardPosselt:
Which makes sense given that you usually want to just download the latest version :)

janibonnevier:
Non-nightly version, right?

BernhardPosselt:
Yes, however we also need to think about how to handle nightly versions. 

Given that they are unstable and potentially insecure and might break your system we should not display them as prominent as the other versions. 

I'll let you think of something ;)

PR #101, opened 2016-07-08
App detail page

janibonnevier:
#80

- Image slider
    - Autoscroll with 8 sec interval, disabled once manual controls are used
    - Currently no indication of where in the slideshow you are
    - No jQuery
- List the latest release for each compatible platform version
    - Nightly releases are ignored
    - Involves some new methods in models.py, some extracted from an API view
    - ~~Platform versions are currently sorted alpabetically, meaning 10<9. That should probably be fixed.~~
- Markdown rendered in description
- "Featured" tag
- Layout and design stuff


nextcloudappstore/core/models.py
```diff
@@ -118,6 +159,12 @@ def can_delete(self, user: User) -> bool:
     def __str__(self) -> str:
         return '%s %s' % (self.app, self.version)
 
+    def is_compatible(self, platform_version):
+        platform_version = Version(platform_version)
+        if platform_version in Spec(self.platform_version_spec):
```
BernhardPosselt: What about simply doing: return platform_version in Spec(self.platform_version_spec)

janibonnevier: Yes, of course!

nextcloudappstore/core/models.py
```diff
@@ -70,6 +84,33 @@ def can_update(self, user: User) -> bool:
...
+
+        Ignores nightly releases.
+        """
+
+        all_latest = {}
+
+        def latest_non_nightly(releases):
+            latest = None
+            for release in releases:
+                if (latest is None or release.version > latest.version) \
```
janibonnevier: I realize now that this is a really stupid way to compare versions.

nextcloudappstore/core/templatetags/dict_sort.py
```diff
@@ -0,0 +1,12 @@
+from django import template
+
+register = template.Library()
+
+
+@register.filter()
+def dict_sort(value, arg):
+    reverse = True if arg == 'desc' else False
+    ret_list = []
+    for key in sorted(value.keys(), reverse=reverse):
```
janibonnevier: Fixed so that platform versions are ordered correctly on the app detail page.

nextcloudappstore/settings/base.py
```diff
@@ -198,3 +198,5 @@
 
 LOGIN_REDIRECT_URL = 'home'
 LOGIN_URL = 'account_login'
+
+PLATFORM_VERSIONS = []
```
BernhardPosselt: We should add '9.0' by default

nextcloudappstore/core/views.py
```diff
@@ -14,6 +15,9 @@ class AppDetailView(DetailView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['categories'] = Category.objects.all()
+        context['latest_releases_by_platform_v'] = \
+            self.object.latest_releases_by_platform_v()
+        context['desc_markdown'] = markdown(self.object.description)
```
BernhardPosselt: We need to make sure that \[link name](javascript:alert(1)) is not being turned into \<a href="javascript:alert(1)">

Can you check if that's the case?

nextcloudappstore/core/tests/test_compatibility.py
```diff
@@ -0,0 +1,72 @@
...
+from nextcloudappstore.core.models import App, AppRelease
+
+
+class CompatibilityTest(TestCase):
+    def setUp(self):
+        self.user = get_user_model().objects.create_user(username='test',
+                                                         password='test',
+                                                         email='test@test.com')
+
+        self.platform_versions = ['9.0.0', '9.0.5', '9.1.3', '10.0.0']
```
BernhardPosselt: we probably don't want to mark up platform versions in 3 characters.

nextcloudappstore/core/static/assets/js/img-slider.js
```diff
@@ -0,0 +1,77 @@
+
+var IMG_SLIDER_CONST = {
```
BernhardPosselt: This should be passed into the constructor :)

nextcloudappstore/core/static/assets/js/img-slider.js
```diff
@@ -0,0 +1,77 @@
...
+    AUTOSCROLL_INTERVAL: 8000 // ms
+};
+
+
+function ImageSlider(element, nextBtnElement, prevBtnElement) {
```
BernhardPosselt: This is essentially a class, can you migrate it?

nextcloudappstore/core/static/assets/js/img-slider.js
```diff
@@ -0,0 +1,77 @@
...
+        clearInterval(slider.autoScroll);
+    });
+    prevBtnElement.addEventListener('click', function() {
+        slider.increment(-1);
+        clearInterval(slider.autoScroll);
+    });
+}
+
+ImageSlider.prototype.setSlide = function(slide) {
+    var imgSpacing = 4;
```
BernhardPosselt: It would be awesome if we could go block scoping only. Function scoping creates all sorts of issues, especially with async closures

nextcloudappstore/core/models.py
```diff
@@ -118,6 +159,12 @@ def can_delete(self, user: User) -> bool:
     def __str__(self) -> str:
         return '%s %s' % (self.app, self.version)
 
+    def is_compatible(self, platform_version):
+        platform_version = Version(platform_version)
```
BernhardPosselt: for platform_version we want to use the pad_min_version function here https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/versioning.py#L25

The reasoning behind this is that a user does not want to see Nextcloud patch versions when downloading archives (e.g. he does not want 9.0.0, he wants 9.0).

nextcloudappstore/core/models.py
```diff
@@ -20,6 +21,19 @@ def search(self, terms, lang):
...
         return queryset.filter(query)
 
+    def get_compatible(self, platform_version):
+        apps = App.objects.prefetch_related('translations', 'screenshots',
+                                            'releases', 'releases__databases',
+                                            'releases__php_extensions').all()
+
+        def app_filter(app):
+            for release in app.releases.all():
+                if release.is_compatible(platform_version):
```
BernhardPosselt: return release.is_compatible(platform_version)

nextcloudappstore/settings/base.py
janibonnevier: I meant to ask about that. Currently it only works for full semver numbers, like 9.0.0, but we might want to support shorter/broader platform numbers.

I resorted to adding some version numbers to this in `settings/development.py`. Do I put the default in `settings/production.py`?

nextcloudappstore/core/models.py
BernhardPosselt: I'd filter out all nightly versions, sort the list by version and return the first element or None, e.g.:

```python
releases = filter(lambda r: not r.endswith('-nightly')
if len(releases) == 0:
    return None

# you may need to read up on the sorted syntax, just a guess ;)
sorted_releases = sorted(releases, lambda r: Version(r))
return sorted_releases[0]
```

```

File: nextcloudappstore/core/views.py (hunk 5 above)
Comment (janibonnevier) [inline nextcloudappstore/core/views.py]: While the code did not run, it did indeed become `<a href="javascript:alert(1)">link name</a>`.

File: nextcloudappstore/core/models.py [hunk 12]
```diff
@@ -20,6 +21,19 @@ def search(self, terms, lang):
         query = reduce(lambda x, y: x & y, predicates, Q())
         return queryset.filter(query)
 
+    def get_compatible(self, platform_version):
```
BernhardPosselt: We also need to think about what do with releases that depend on a minor platform version, e.g.

```xml
<owncloud min-version="9.0.1" max-version="9.0"/>
```
This release should appear under the 9.0 tab in the web interface, but in the API we get the full current platform version (e.g. 9.0.2) and therefore should not be included in the result.

nextcloudappstore/settings/base.py
BernhardPosselt: Right, so the basic issue is that ownCloud has a weird versioning schema, as in: they don't follow semver or any other commonly used version schema.

As a user you want to see download buttons for your base version, e.g. Releases compatible with 9.0, Relases compatible with 9.1, etc

This is the thing that needs to be rendered on the page.

BernhardPosselt: We've got utility functions for padding min and max versions btw :)

nextcloudappstore/core/views.py
BernhardPosselt: Still not fixed hm :/

We need to figure out how to prevent this or go for a JS solution which disallows that

nextcloudappstore/core/models.py
janibonnevier: No, it needs to check all releases for a given app. This change would mean that whatever release it checks first determines whether the app as a whole is compatible with a platform version.

nextcloudappstore/core/models.py
BernhardPosselt: BTW using a map to turn it into a list of versions and then using Max function would even be better

nextcloudappstore/core/views.py
janibonnevier: [bleach](https://pypi.python.org/pypi/bleach) seems to do the job. Shall we use it?

BernhardPosselt: I don't think it's needed, we just need a proper markdown lib. I'll search something this evening

nextcloudappstore/core/models.py
BernhardPosselt: ah right, its indented two levels

PR #109, opened 2016-07-10
Add update script

BernhardPosselt:
@LukasReschke this should allow you to update the app store by simply running:

    bash scripts/maintenance/update.sh


adsworth:
looks good to me.

PR #110, opened 2016-07-10
Fix #94

### LINKED ISSUE #94: App title and description must be available in English
Both can be localized however we don't expect everything to be translated. Because we use English as fallback if a language is not present we should require the English contents in the info.xml

BernhardPosselt:
@janibonnevier @adsworth


janibonnevier:
Looks good. :+1:

Issue #105, opened 2016-07-09
App Store should be mentioned in the logo


BernhardPosselt:
We need to get the App Store in somewhere, maybe the logo?

@jancborchardt @adsworth @janibonnevier

jancborchardt:
Yeah, the layout should be like in Nextcloud itself, just put »App store« (small s, we only capitalize the first letter ;) right next to the logo.

janibonnevier:
Should the "App store" text be translatable?

janibonnevier:
Nvm, I'll just do it.

BernhardPosselt:
Nah, app store is a name (and should therefore be all caps but whatever :D)

PR #122, opened 2016-07-11
App store text in banner

janibonnevier:
#105 

- Add "App store" text to banner.
- Redo banner CSS.
- Set banner `background-size: cover`.
- De-capitalize "Store" in `<title>`.
- Add a margin at bottom of sidebar because of mobile.


BernhardPosselt:
What If we made the banner bg blue and remove the bg image and then go for 

"Logo" Nextcloud App Store

Where logo is the simple logo without text, similar to your Nextcloud instance

Also we should capitalize it, otherwise we look like people who don't know English.

BernhardPosselt:
We could also cut the banner height in half if we go for the blue :)

janibonnevier:
My arguments for keeping the background image are

- It's consistent with nextcloud.com.
- It adds some depth to an otherwise very flat design, if you think that's a good thing. 
- I like it. 

Your call. :)

BernhardPosselt:
I dont think we can make the current logo work with the 3 words also, also the 2 font families don't match at all. We need to have the heading in one font family.

Keeping the three words under the logo is not an option, it will probably look kinda weird, so my idea would be to have all of them in one line. This creates a very wide banner so I'd make the height shorter. This in turn will hide most of the bg image. We can take a look if it looks ok, otherwise I'd just scrap it and go for the Nextcloud look (not nextcloud.com but the actual software)

BernhardPosselt:
cloud

BernhardPosselt:
Kinda like this (you can also try the bg image ;))

janibonnevier:
Yes, I agree about the text. What font is that? We should probably use it elsewhere too.

BernhardPosselt:
Open Sans afaik

janibonnevier:
Try it out without the bg image too. I added the bg color as fallback regardless.

nextcloudappstore/core/static/assets/css/fonts.css
```diff
@@ -0,0 +1,120 @@
+/* Generated by Font Squirrel (https://www.fontsquirrel.com) on July 12, 2016 */
+
+
+
+@font-face {
+    font-family: 'open_sanssemibold_italic';
```
jancborchardt: We never ever use italic faces anywhere. :) Only Light and Semibold. Emphasis (or de-emphasis) can be achieved by opacity / greyscale, which looks a lot better.

nextcloudappstore/core/static/assets/css/fonts.css
```diff
@@ -0,0 +1,120 @@
...
+    font-weight: normal;
+    font-style: normal;
+
+}
+
+
+
+
+@font-face {
+    font-family: 'open_sansextrabold';
```
jancborchardt: Same for these – we only use Light and Semibold. No need for Extra Bold etc. :)

jancborchardt:
We should keep it consistent with the website header background as in the future we should have it more closely integrated there. Also as @janibonnevier says it gives it some depth and just looks a lot nicer.

@BernhardPosselt the text in the header should be [Logo] App store. Just like in Nextcloud itself for the apps. Of course the logo should have `alt="Nextcloud"`.

Cool work @janibonnevier btw, great to have you on board! :) I also added you to the Nextcloud design team, if you want to contribute to the server for example.

nextcloudappstore/core/static/assets/css/fonts.css
jancborchardt: You can just copy the fonts.css file from the server: https://github.com/nextcloud/server/blob/master/core/css/fonts.css

nextcloudappstore/core/templates/base.html
```diff
@@ -9,6 +9,7 @@
     <meta name="author" content="">
     <link rel="shortcut icon" href="{% static 'assets/img/favicon.png' %}"/>
     <title>Nextcloud App Store{% block head-title %}{% endblock %}</title>
```
jancborchardt: The title should be »App store - Nextcloud«. The more specific thing needs to go first so it works properly when you have a bunch of tabs open.

janibonnevier:
@jancborchardt, thanks! When you say [Logo], you mean the full logo that says "Nextcloud", right?

jancborchardt:
I’m not entirely sure about that one. @BernhardPosselt above preferred the logo without the text, but having »Nextcloud App store« in the title is a bit much maybe. And then with the text it’s more like the website, so I would say we should go with that first.

nextcloudappstore/core/templates/base.html
janibonnevier: I certainly see the point, but the question arises whether the ordering of "App store - Nextcloud - [Appstore subpage]" makes any sense. Perhaps in that case we should go for "[Appstore subpage] - App store - Nextcloud"? @BernhardPosselt

BernhardPosselt:
I'd say lets merge it for now and wait with the design review until we've got something almost feature complete like a beta. We will have a better overview how everything fits together and I think a lot of design details will have changed :)

As far as the PR goes: It basically solves the issue. I don't think we should lose ourselves in  small details which were not specified in the issue since its very frustrating for contributors and stalls people working on it.  

@jancborchardt I'd prefer if we could postpone the changes until we've got the profile and releases page. Then let's document and discuss all the design changes in one or more tickets so people working on it have clear guidelines :)

nextcloudappstore/core/templates/base.html
BernhardPosselt: Agree with @janibonnevier we want a hirarchy and I'd keep the title as is since its the top level element :)

BernhardPosselt:
Fine to merge from my POV :) :+1:

nextcloudappstore/core/templates/base.html
adsworth: I'd also go for [subpage] - App store - Nextcloud. I hate pages where the important part of the title is last. That makes finding the right tab a real pain. :)

Other than that everthing looks fine for me.

janibonnevier:
Found a little glitch, hold on.

nextcloudappstore/core/templates/base.html
BernhardPosselt: Ah, good point. It becomes a pain indeed when tabs are cramped

BernhardPosselt: I'll move this to a separate ticket

adsworth: In Firefox it is a pain even if tabs are not cramped, as Firefox has a max tab width of around 22 characters or so.

BernhardPosselt: Moved to https://github.com/nextcloud/appstore/issues/125 since it touches more places than this issue

jancborchardt:
Cool, sounds good

nextcloudappstore/core/templates/base.html
jancborchardt: Yeah, just for the record – of course also sub page titles should go first then. The most specific element should always go left. As I said in the OP:

> The more specific thing needs to go first so it works properly when you have a bunch of tabs open.

Issue #108, opened 2016-07-09
My apps filter


BernhardPosselt:
There should be a filter in the category list which should be called "My apps". Clicking on it will display all apps that you own and co-maintain.

PR #124, opened 2016-07-12
Filter apps by maintainer in app list

janibonnevier:
#108 

Incl. filter checkbox that sets `?maintainer=[current user username]`.

### DIFF
--- a/nextcloudappstore/core/views.py
+++ b/nextcloudappstore/core/views.py
@@ -64,7 +64,7 @@ class CategoryAppListView(ListView):
                 queryset = queryset.filter(Q(owner=user) |
                                            Q(co_maintainers=user))
             except ObjectDoesNotExist:
-                queryset = None
+                queryset = []
         return queryset
 
     def get_context_data(self, **kwargs):
--- a/nextcloudappstore/core/views.py
+++ b/nextcloudappstore/core/views.py
@@ -64,7 +64,7 @@ class CategoryAppListView(ListView):
                 queryset = queryset.filter(Q(owner=user) |
                                            Q(co_maintainers=user))
             except ObjectDoesNotExist:
-                queryset = []
+                return []
         return queryset
 
     def get_context_data(self, **kwargs):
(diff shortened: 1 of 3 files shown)


nextcloudappstore/core/views.py
```diff
@@ -64,6 +75,8 @@ def get_context_data(self, **kwargs):
             context['current_category'] = Category.objects.get(id=category_id)
         if self.search_terms:
             context['search_query'] = ' '.join(self.search_terms)
+        context['auth_user'] = self.request.user
```
BernhardPosselt: this should already be available in the templates same for the following line

janibonnevier: Couldn't find anything when I looked, but I'll look some more.

BernhardPosselt: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/templates/nav.html#L46

nextcloudappstore/core/views.py
```diff
@@ -54,6 +58,13 @@ def get_queryset(self):
             queryset = queryset.filter(categories__id=category_id)
         if featured == "true":
             queryset = queryset.filter(featured=True)
+        if maintainer:
+            try:
+                user = User.objects.get_by_natural_key(maintainer)
+                queryset = queryset.filter(Q(owner=user) |
+                                           Q(co_maintainers=user))
+            except ObjectDoesNotExist:
+                return []
```
janibonnevier: At least the result is as desired in this case, and I've previously seen that a list can sometimes take the place of a queryset. Would you rather have something like `return QuerySet()` or `queryset = queryset.filter(owner__username__exact=maintainer)`?

janibonnevier: Hey! There's `none()`, I'll use that of course!

nextcloudappstore/core/views.py
```diff
@@ -54,6 +58,13 @@ def get_queryset(self):
             queryset = queryset.filter(categories__id=category_id)
         if featured == "true":
```
janibonnevier: Because if we simply check if `featured` is truthy, `?featured=false` becomes "show only featured apps", because the string `"false"` is truthy. `maintainer` on the other hand may be any username, so we just check if one was provided.

BernhardPosselt: ah right

BernhardPosselt:
Looks good

Issue #79, opened 2016-07-01
App release page


BernhardPosselt:
Instead of listing all app versions in a long list a typical user wants to lookup which version runs on his platform. Therefore the releases page should order releases based on their supported platform version (Think of it like Nextcloud 9.0, Nextcloud 9.1 etc). 

Each release should list the database, shell commands, extensions, licenses, php and shell dependencies, checksum, download link below. Also a relative date should be shown for last_modified property (e.g. updated 8 months ago). Hovering over the relative date should show the actual date (use the title attribute).

All infos should be collapsed (click to expand) except the latest one for the appropriate platform.

We should also think about a way to communicate support timeframes to a user. It makes no sense to list all possible releases equally if only the latest one receives security updates. As a user I want to know if the latest version for my platform will receive security updates.

PR #126, opened 2016-07-13
App releases page

janibonnevier:
Not quite everything from #79 is in place, such as relative time in the "Updated" field, but I put this up here for review anyway.

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -86,8 +86,18 @@ class App(TranslatableModel):
     def can_delete(self, user: User) -> bool:
         return self.owner == user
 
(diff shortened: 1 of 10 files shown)


BernhardPosselt:
Well done. I think the following things should probably be added:

* The latest release of each platform should be expanded by default
* If no shell commands or PHP extensions are required (size 0), I'd hide the section
* If a version spec is \* I'd hide it. I don't think a lot of users know wildcards and if all versions are supported its best to omit it i think

BernhardPosselt:
Example for hiding the version spec:
```
SimpleXML
curl
iconv
libxml >=2.7.8
```

BernhardPosselt:
:+1:

PR #134, opened 2016-07-16
Remove CSP on the signup page

BernhardPosselt:
I've completely removed CSP from the sign up page which should be ok since its static content (i guess) and it breaks recaptcha otherwise

@adsworth @LukasReschke


BernhardPosselt:
Better hardening is in this ticket https://github.com/nextcloud/appstore/issues/136

PR should be "secure enough" for now :)

LukasReschke:
👍

Issue #99, opened 2016-07-05
Cumulative GET paremeters on app list page


BernhardPosselt:
The GET parameters should be cumulative meaning:
* If I search a term and I change the category, the search term should be kept but the category filtering should change
* If I change the sorting after a search, it should sort the search result appropriately
* If I change the category the ordering should be kept

Solution for this is probably (not sure, just a guess) a template tag.

janibonnevier:
> If I search a term and I change the category, the search term should be kept but the category filtering should change

I myself would expect the search to be discarded if I click on a category, because in my opinion they are conceptually not filters that operate on the currently visible list of apps, but links to other lists.

BernhardPosselt:
So your take on it would be to only update the GET variables inside a category and resetting them when choosing another category correct?

janibonnevier:
Yes, and I think it is because the list of categories looks like navigation, so by clicking there I expect to navigate away from any search results. I'm not so sure about ordering and the "Featured" and "My apps" filters, however...

BernhardPosselt:
Fine by me :)

Do the filters in a category work in a cumulative fashion yet? Also including search?

janibonnevier:
I'm working on it. Basically it works but I'll try to make the solution somewhat pretty too.

PR #137, opened 2016-07-17
Preserve GET params in app list view

janibonnevier:
#99 

- Preserve GET params according to the table below.
- Clearly indicate to the user when they are browsing search results.

P=Preserve, D=Discard

Change | Search | Filters | Ordering | Category
--- | --- | --- | --- | ---
Search | - | D | P | P
Filters | P | - | P | P
Ordering | P | P | - | P
Category | D | P | P | -

In short, everything is preserved except search when changing category and filters when searching. The former provides a natural way to discard search results (it would be rather difficult otherwise). The latter is because the search box gives no indication that the search is limited by the filters. I think that's something for a future advanced search. However, even now it is possible to apply filters to search results once they are presented to you.

### DIFF
--- a/nextcloudappstore/core/templates/app/list.html
+++ b/nextcloudappstore/core/templates/app/list.html
@@ -12,22 +12,22 @@
 {% if search_query %}
     <div class="col-lg-6">
         {% if search_query and current_category %}
-            <h2 class="search-results-title">{% blocktrans with cat=current_category q=search_query%}Search results for "{{ q }}" in {{ cat }}{% endblocktrans %}</h2>
+            <h2 class="search-results-title">{% blocktrans with category=current_category query=search_query %}Search results for "{{ query }}" in {{ category }}{% endblocktrans %}</h2>
         {% elif search_query %}
-            <h2 class="search-results-title">{% blocktrans with q=search_query%}Search results for "{{ q }}"{% endblocktrans %}</h2>
+            <h2 class="search-results-title">{% blocktrans with query=search_query %}Search results for "{{ query }}"{% endblocktrans %}</h2>
         {% endif %}
     </div>
 {% endif %}
 
 <div class="row">
     <ul class="nav-tabs nav sorting">
-        <li role="presentation" class="{% if not request.GET.order_by or request.GET.order_by == 'last_modified' %}active{% endif %}"><a href="{{ request.url }}?order_by=last_modified&ordering=desc{% if get_arg_url_strings.search_filters %}&{{ get_arg_url_strings.search_filters }}{% endif %}">{% trans 'Latest' %}</a></li>
+        <li role="presentation" class="{% if not request.GET.order_by or request.GET.order_by == 'last_modified' %}active{% endif %}"><a href="{{ request.url }}?order_by=last_modified&amp;ordering=desc{% if url_params.search_filters %}&amp;{{ url_params.search_filters }}{% endif %}">{% trans 'Latest' %}</a></li>
         {% if request.GET.order_by == 'name' and request.GET.ordering == 'asc' %}
-        <li role="presentation" class="{% if request.GET.order_by == 'name' %}active{% endif %}"><a href="{{ request.url }}?order_by=name&ordering=desc{% if get_arg_url_strings.search_filters %}&{{ get_arg_url_strings.search_filters }}{% endif %}"><span class="glyphicon glyphicon-triangle-bottom" aria-hidden="true"></span> {% trans 'Alphabetical' %}</a></li>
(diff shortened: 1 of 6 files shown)


nextcloudappstore/core/templates/app/list.html
```diff
@@ -8,15 +8,26 @@
 {% endblock %}
 
 {% block apps %}
+
+{% if search_query %}
+    <div class="col-lg-6">
+        {% if search_query and current_category %}
+            <h2 class="search-results-title">{% blocktrans with cat=current_category q=search_query%}Search results for "{{ q }}" in {{ cat }}{% endblocktrans %}</h2>
```
BernhardPosselt: probably better to spell out query and category variables inside the translation so their meaning is obvious to translators

nextcloudappstore/core/templates/app/list.html
```diff
@@ -8,15 +8,26 @@
...
+        {% elif search_query %}
+            <h2 class="search-results-title">{% blocktrans with q=search_query%}Search results for "{{ q }}"{% endblocktrans %}</h2>
```
BernhardPosselt: Same here

nextcloudappstore/core/templates/app/list.html
```diff
@@ -8,15 +8,26 @@
...
+        {% endif %}
+    </div>
+{% endif %}
+
 <div class="row">
     <ul class="nav-tabs nav sorting">
-        <li role="presentation" class="{% if not request.GET.order_by or request.GET.order_by == 'last_modified' %}active{% endif %}"><a href="{{ request.url  }}?order_by=last_modified&ordering=desc">{% trans 'Latest' %}</a></li>
+        <li role="presentation" class="{% if not request.GET.order_by or request.GET.order_by == 'last_modified' %}active{% endif %}"><a href="{{ request.url }}?order_by=last_modified&ordering=desc{% if get_arg_url_strings.search_filters %}&{{ get_arg_url_strings.search_filters }}{% endif %}">{% trans 'Latest' %}</a></li>
         {% if request.GET.order_by == 'name' and request.GET.ordering == 'asc' %}
-        <li role="presentation" class="{% if request.GET.order_by == 'name' %}active{% endif %}"><a href="{{ request.url  }}?order_by=name&ordering=desc"><span class="glyphicon glyphicon-triangle-bottom" aria-hidden="true"></span> {% trans 'Alphabetical' %}</a></li>
```
BernhardPosselt: &amp; I think instead of &

nextcloudappstore/core/views.py
```diff
@@ -89,8 +91,46 @@ def get_context_data(self, **kwargs):
             context['current_category'] = Category.objects.get(id=category_id)
         if self.search_terms:
             context['search_query'] = ' '.join(self.search_terms)
+        context['get_arg_url_strings'] = self.get_arg_url_strings
```
BernhardPosselt: url_params would probably be shorter and more readable :)

nextcloudappstore/core/views.py
```diff
@@ -89,8 +91,46 @@ def get_context_data(self, **kwargs):
...
+            'search_filters': self._combine_arg_strings(search, filters),
+        }
+
+    def _get_args_as_url_string(self, *arg_names):
+        args = map(lambda arg: (arg, self.request.GET.get(arg, '')), arg_names)
+        filtered_dict = dict(filter(lambda a: a[1], args))
+        return urlencode(filtered_dict)
+
+    def _combine_arg_strings(self, *strings):
+        return reduce(lambda str, add: str+'&'+add, filter(None, strings), '')\
```
BernhardPosselt: maybe '&'.join(strings) is better here (reduce is basically join for functions)

nextcloudappstore/core/views.py
```diff
@@ -89,8 +91,46 @@ def get_context_data(self, **kwargs):
...
+            'search': search,
+            'filters': filters,
+            'ordering': ordering,
+            'filters_ordering': self._combine_arg_strings(filters, ordering),
+            'search_filters': self._combine_arg_strings(search, filters),
+        }
+
+    def _get_args_as_url_string(self, *arg_names):
+        args = map(lambda arg: (arg, self.request.GET.get(arg, '')), arg_names)
+        filtered_dict = dict(filter(lambda a: a[1], args))
```
BernhardPosselt: may be more readable to move the filter out of the dict function and call the result present_args

nextcloudappstore/core/views.py
```diff
@@ -89,8 +91,46 @@ def get_context_data(self, **kwargs):
...
+        }
+
+    def _url_params_str(self, *params):
+        args = map(lambda param: (param, self.request.GET.get(param, '')),
+                   params)
+        present_args = filter(lambda a: a[1], args)
+        return urlencode(dict(present_args))
+
+    def _join_url_params_strs(self, *strings):
+        return '&'.join(strings).lstrip('&')
```
BernhardPosselt: whats the reason for using lstrip here?

BernhardPosselt:
:+1:

PR #142, opened 2016-07-19
Fix #88

### LINKED ISSUE #88: Outgoing links should add rel="noreferrer"
This applies to links parsed from the info.xml as well as the app's description markdown

janibonnevier:
#88 

- App resource links.
- Links in app description.
- Release download links on app detail and release pages.

### DIFF
--- a/nextcloudappstore/core/static/assets/js/app-detail.js
+++ b/nextcloudappstore/core/static/assets/js/app-detail.js
@@ -277,7 +277,14 @@
     fetch(appDescriptionUrl.content).then((response) => {
         return response.text()
     }).then((description) => {
+        let mdDesc = md.render(description);
+
+        // Add noopener and noreferrer to all links in the description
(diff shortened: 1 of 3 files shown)


nextcloudappstore/core/static/assets/js/app-detail.js
```diff
@@ -277,7 +277,14 @@
     fetch(appDescriptionUrl.content).then((response) => {
         return response.text()
     }).then((description) => {
+        let mdDesc = md.render(description);
+
+        // Add noopener and noreferrer to all links in the description
```
janibonnevier: Hmm, maybe this bit should be extracted to a function? What do you think?

BernhardPosselt: yes

BernhardPosselt:
:+1:

PR #147, opened 2016-07-20
User account settings pages

### LINKED ISSUE #39: add a user profile
add a user profile with:
-list of apps and their releases. Allow deleting of apps and releases, but no editing. Updates should be done via the API.
- allow changing of password.

Anything else?

adsworth:
other than password change (which is a allauth view) the views don't actually do anything. :)

will fix #39


nextcloudappstore/core/templates/account/password_change.html
```diff
@@ -0,0 +1,47 @@
...
+
+    <h1>{% trans "Change Password" %}</h1>
+
+    <form method="POST" action="{% url 'account_change_password' %}" class="password_change">
+        {% csrf_token %}
+        {% if form.non_field_errors  %}
+            {% for error in form.non_field_errors %}
+                <span class="glyphicon glyphicon-remove form-control-feedback" aria-hidden="true"></span>
+                <span class="sr-only">(error)</span>
+                <p class="text-danger">{{ error|escape }}</p>
```
BernhardPosselt: should be escaped by default, right?

nextcloudappstore/core/templates/account/password_change.html
```diff
@@ -0,0 +1,47 @@
...
+            {% if field.errors %}
+                {% for error in field.errors %}
+                    <span class="glyphicon glyphicon-remove form-control-feedback" aria-hidden="true"></span>
+                    <span class="sr-only">(error)</span>
+                    <p class="text-danger">{{ error|escape }}</p>
+                {% endfor %}
+            {% endif %}
+
+            {% if field.help_text %}
+                <p class="help-block">{{ field.help_text|safe }}</p>
```
BernhardPosselt: I dont think we want safe here

nextcloudappstore/core/templates/account/password_change.html
```diff
@@ -0,0 +1,47 @@
...
+            {% endif %}
+
+        </div>
+        {% endfor %}
+        <p><button class="btn btn-primary btn-block" type="submit">{% trans "Change Password" %}</button></p>
+        {% if redirect_field_value %}
+            <input type="hidden" name="{{ redirect_field_name }}" value="{{ redirect_field_value }}" />
```
BernhardPosselt: Are we sure that this can not be controled by an attacker? Is this a static value or will it be validated?

nextcloudappstore/core/templates/user/change_name.html
```diff
@@ -0,0 +1,10 @@
+{% extends "user/base.html" %}
+{% load css_class %}
+{% load i18n %}
+{% load staticfiles %}
+
+{% block head-title %}{% trans "Change name" %} - {% endblock %}
+
+{% block profile-content %}
+    <h1>Change name</h1>
```
BernhardPosselt: l10n

nextcloudappstore/core/templates/user/api_token.html
```diff
@@ -0,0 +1,10 @@
+{% extends "user/base.html" %}
+{% load css_class %}
+{% load i18n %}
+{% load staticfiles %}
+
+{% block head-title %}{% trans "API token" %} - {% endblock %}
+
+{% block profile-content %}
+    <h1>API token</h1>
```
BernhardPosselt: l10n

nextcloudappstore/core/templates/user/profile.html
```diff
@@ -0,0 +1,17 @@
...
+
+{% block head-title %}{% trans "Profile" %} - {% endblock %}
+
+{% block profile-content %}
+    <h1>{% trans "My apps" %}</h1>
+    <ul class="nav nav-pills nav-stacked">
+    {% for obj in object_list %}
+        <li><a href="{{ obj.get_absolute_url }}">{{ obj }}</a></li>
+    {% empty %}
+        <li>No Apps</li>
```
BernhardPosselt: l10n

nextcloudappstore/core/user/views.py
```diff
@@ -1 +1,27 @@
...
+
+
+class ProfileView(LoginRequiredMixin, ListView):
+    """Display the users profile"""
+    template_name = 'user/profile.html'
+    model = App
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+        return qs.filter(owner=self.request.user)
```
BernhardPosselt: I think this should be a DetailView and should use the User model as queryset

nextcloudappstore/core/templates/account/password_change.html
adsworth: It's what we use in the login form and help_text can't be passed by the user, it's the value of the models field help_text fom the source.

nextcloudappstore/core/templates/account/password_change.html
adsworth: in this case we might not need a redirect_field. The form is basically a copy & paste of the login form which validates the redirect.

nextcloudappstore/core/templates/account/password_change.html
BernhardPosselt: So its not translated, right?

BernhardPosselt: Ah right, it is :)

However I'd still remove safe, we dont want to add html to help_text

adsworth:
for me the main questions still is: Do we keep it like it currently is one view for each action and the main profile page displays the user apps. Or do we make one big profile page with tabs/accordion that contains all the forms, name change, token and password change and then use different views to actually make the change and then redirect back to the main profile page?

BernhardPosselt:
I think it looks fine like that. 

Checked out the branch atm and noticed some more things :)
* Token is not displayed
* The currently active navigation entries should be marked with the css active class
* I'd remove the my apps section, its a filter in the apps category already

BernhardPosselt:
I'd create the following sections:
* General: Includes Token and form for changing the name and email
* Credentials: Includes the password reset form and the token (with a button to generate a new token)

adsworth:
and what do we display instead of the apps on the main profile page?

BernhardPosselt:
The profile form

BernhardPosselt:
BTW, I think GitHub has a very neat profile page :)

adsworth:
and they display my repos when I click on "Your profile" :)

adsworth:
so profile forms == user name change form?

BernhardPosselt:
Click on edit profile ;D

BernhardPosselt:
Yes, user, email (if possible) for now

nextcloudappstore/core/user/views.py
```diff
@@ -1 +1,89 @@
...
+    template_name = 'user/api_token.html'
+
+    def get_context_data(self, **kwargs):
+        context = super(APITokenView, self).get_context_data(**kwargs)
+        context['acc_page'] = 'api_token'
+        if self.request.user.is_authenticated():
+            token = Token.objects.get_or_create(user=self.request.user)[0].key
+        else:
+            token = ''
+        context['token'] = token
```
janibonnevier: Unsure if this is acceptable.

nextcloudappstore/core/user/views.py
```diff
@@ -1 +1,89 @@
...
+        return context
+
+    def post(self, request):
+        if self.request.user.is_authenticated():
+            try:
+                Token.objects.get(user=self.request.user).delete()
+            except:
+                pass
+            new = Token.objects.create(user=self.request.user)
+            new.save()
```
janibonnevier: Unsure whether this is a good way to do this.

janibonnevier:
Apparently I did not read the previous comments thoroughly. Oh well.

nextcloudappstore/core/templates/nav.html
```diff
@@ -52,21 +52,24 @@
...
+                    <li class="nav-link">
+                        <a href="{% url 'admin:index' %}">{% trans 'Admin' %}</a>
+                    </li>
                 {% endif %}
                 {% if request.user.is_authenticated %}
-                <li class="nav-link">
-                    <a href="{% url 'account_logout' %}">{% trans 'Log out' %}</a>
-                </li>
+                    <li class="nav-link">
+                        <a href="{% url 'user-account' %}">Account</a>
```
BernhardPosselt: trans missing :)

nextcloudappstore/core/templates/user/api_token.html
```diff
@@ -0,0 +1,20 @@
...
+{% load i18n %}
+{% load staticfiles %}
+
+{% block head-title %}{% trans "API token" %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "API Token" %}</h1>
+    <section>
+        <p>Your API token is:</p>
+        <p><code>{{ token }}</code></p>
```
BernhardPosselt: We need to query the token via JS (like on the upload page) because of BREACH

nextcloudappstore/core/templates/user/api_token.html
```diff
@@ -0,0 +1,20 @@
...
+    </section>
+    <section>
+        <form action="{{ request.url }}" method="post">
+            {% csrf_token %}
+            <button class="btn btn-primary btn-block" type="submit">Regenerate token</button>
```
BernhardPosselt: trans

nextcloudappstore/core/templates/user/api_token.html
```diff
@@ -0,0 +1,20 @@
...
+{% load css_class %}
+{% load i18n %}
+{% load staticfiles %}
+
+{% block head-title %}{% trans "API token" %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "API Token" %}</h1>
+    <section>
+        <p>Your API token is:</p>
```
BernhardPosselt: trans

nextcloudappstore/core/templates/user/api_token.html
BernhardPosselt: Needs renewal needs to happen via JS

nextcloudappstore/core/user/views.py
```diff
@@ -1 +1,72 @@
...
+
+
+class APITokenView(LoginRequiredMixin, TemplateView):
+    """Display the user's API Token."""
+    template_name = 'user/api_token.html'
+
+    def get_context_data(self, **kwargs):
+        context = super(APITokenView, self).get_context_data(**kwargs)
+        context['acc_page'] = 'api_token'
+        if self.request.user.is_authenticated():
```
BernhardPosselt: it would be interesting if LoginRequiredMixin already takes care of that which would mean you will never run into the else block, same for the next method

nextcloudappstore/core/user/views.py
```diff
@@ -1 +1,72 @@
...
+        context['token'] = token
+        return context
+
+    def post(self, request):
+        if self.request.user.is_authenticated():
+            try:
+                Token.objects.get(user=self.request.user).delete()
+            except:
+                pass
+            new = Token.objects.create(user=self.request.user)
```
BernhardPosselt: Unsure, but I think the create already takes care of saving it, so the next save might be redundant

janibonnevier: Correct.

nextcloudappstore/core/api/v1/views.py
```diff
@@ -167,3 +169,36 @@ def post(self, request, *args, **kwargs):
...
+    throttle_classes = (PostThrottle,)
+    permission_classes = (IsAuthenticated,)
+    parser_classes = (parsers.FormParser,
+                      parsers.MultiPartParser,
+                      parsers.JSONParser,)
+    renderer_classes = (renderers.JSONRenderer,)
+    serializer_class = AuthTokenSerializer
+
+    def post(self, request, *args, **kwargs):
+        if request.user.is_authenticated():
```
BernhardPosselt: No need to do these is_authenticated checks because permission_classes = (IsAuthenticated,) and if the user is not authenticated, it will be null and you cant update it

nextcloudappstore/core/static/assets/js/user-api-token.js
```diff
@@ -0,0 +1,100 @@
...
+            }
+        );
+        return fetch(request).then(global.convertResponse);
+    };
+
+
+    function fetchToken(csrf) {
+        return global.fetchAPIToken(csrf).then(
+            (response) => response.token,
+            () => null
```
BernhardPosselt: why not return the promise itself without the do nothing error handler? than the caller can add a catch() method to handle the error

nextcloudappstore/core/static/assets/js/user-api-token.js
```diff
@@ -0,0 +1,100 @@
...
+    }
+
+
+    let form = document.getElementById('user-api-token-form');
+    let csrfEl = document.getElementsByName('csrfmiddlewaretoken')[0];
+    let confirmText = document.getElementById('regen-confirm-text').innerHTML;
+
+    let token;
+    fetchToken(csrfEl.value).then(
+        (retval) => {
```
BernhardPosselt: data reads better than retval

nextcloudappstore/core/static/assets/js/user-api-token.js
```diff
@@ -0,0 +1,100 @@
...
+    function onTokenRegenFailure() {
+        showTokenRegenFailureMessage(true);
+    }
+
+
+    let form = document.getElementById('user-api-token-form');
+    let csrfEl = document.getElementsByName('csrfmiddlewaretoken')[0];
+    let confirmText = document.getElementById('regen-confirm-text').innerHTML;
+
+    let token;
```
BernhardPosselt: I wouldnt set a "global" token variable. Just update the token and parse it from wherever it was written when you actually need it

nextcloudappstore/core/templates/user/api_token.html
```diff
@@ -0,0 +1,28 @@
...
+{% block head-title %}{% trans "API token" %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "API Token" %}</h1>
+    <section>
+        <div id="regen-success" class="alert alert-success" hidden>{% trans 'New API token generated.' %}</div>
+        <div id="regen-failure" class="alert alert-danger" hidden>{% trans 'Failed to regenerate API token.' %}</div>
+        <span id="regen-confirm-text" hidden>{% trans 'Do you want to regenerate your API token?' %}</span>
+        <p>{% trans 'Your API token is:' %}</p>
+        <p><code id="token" data-error="{% trans 'Failed to fetch token.' %}"></code></p>
```
BernhardPosselt: would prefer the same pattern as above if possible, dont parse the data-error attribute, just display the error div

BernhardPosselt:
BTW, the PR reached a size where I think it cant be reviewed anymore. 

@janibonnevier can you create a new PR and just move the token stuff in there?

BernhardPosselt:
Features from this PR have been implemented or are now in different PRs. Closing this PR. Thanks @adsworth for the head start and @janibonnevier for getting it in :)

PR #150, opened 2016-07-23
Fix #106: Nightlies on app releases page

### LINKED ISSUE #106: Nightly handling on the release page
Currently the download page only displays stable releases. We need a way to also display nightly releases. Nightly releases are unstable and could break your install, so we need to show them less prominently and (probably) display a warning.

janibonnevier:
- Clearly indicate that nightlies are unstable.
- Place nightlies last in each listing.


nextcloudappstore/core/views.py
```diff
@@ -40,10 +41,30 @@ class AppReleasesView(DetailView):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['categories'] = Category.objects.all()
+
+        releases = self.object.releases_by_platform_v()
+        nightlies = self.object.releases_by_platform_v(nightlies=True)
+        versions = set(list(releases.keys()) + list(nightlies.keys()))
```
BernhardPosselt: curious if this would work by using versions = set(zip(release.keys(), nightlies.keys())

nextcloudappstore/core/models.py
```diff
@@ -88,34 +89,52 @@ def can_update(self, user: User) -> bool:
     def can_delete(self, user: User) -> bool:
         return self.owner == user
 
-    def releases_by_platform_v(self):
+    def releases_by_platform_v(self, nightlies=False):
```
BernhardPosselt: probably should look like a bool, e.g. include_nightly or include_nightlies

nextcloudappstore/core/models.py
```diff
@@ -88,34 +89,52 @@ def can_update(self, user: User) -> bool:
...
+
+        :param inclusive: Use inclusive version check (see
+                          AppRelease.is_compatible()).
+        :param nightlies: Return only nightly releases if True, only non-
+                          nightlies otherwise.
+        :return a sorted list of all compatible releases.
+        """
+
         all_releases = self.releases.all()
         return sorted(list(filter(
```
BernhardPosselt: does this also work without list()?

janibonnevier: It does.

nextcloudappstore/core/views.py
janibonnevier: It becomes `{('9.1', '9.1'), ('9.0', '9.0')}` but we want `{'9.0', '9.1'}`.

nextcloudappstore/core/models.py
janibonnevier: It's not inclusion, but exclusive selection, see the docstring. But yes, I'm open to the idea that there might be a better name.

nextcloudappstore/core/models.py
BernhardPosselt: Then lets remove list()

list() forces the iterator to be evaluated which is not only slow but also most of the time unneeded :)

nextcloudappstore/core/views.py
BernhardPosselt: Ah, right :)

nextcloudappstore/core/models.py
janibonnevier: Though I think it is pretty good as it is. :)

BernhardPosselt: exclusive selection? To me it looks like using releases_by_platform(True) would also include nightlies, am i wrong?

BernhardPosselt: Ah, it only returns nightlies. In that case I'd opt for a separate method, e.g.: nighly_releases_by_platform, otherwise its super confusing :)

janibonnevier: Alright :)

PR #151, opened 2016-07-23
Also parse and included authors from the info.xml

### LINKED ISSUE #143: API is missing display of apps owner
Currently we display who is the author of the app in the Nextcloud interface. I think the following  values in https://apps.weasel.rocks/api/v1/platform/9.0.0/apps.json would make sense:

```
"author": {
  "id": "UniqueUsername", // used for generation to profile page if we ever have any
  "displayName": "Hans Muster" // Something we can display in the interface directly
}
```

Thoughts, @BernhardPosselt?

BernhardPosselt:
Fix #143 

@adsworth @LukasReschke @MorrisJobke @janibonnevier


nextcloudappstore/core/models.py
```diff
@@ -123,6 +125,13 @@ def _latest_non_nightly(self, releases):
             return None
 
 
+class AppAuthor(Model):
+    name = CharField(max_length=128, verbose_name=_('Full name'))
+    homepage = URLField(max_length=256, blank=True,
+                        verbose_name=_('Homepage'))
+    mail = EmailField(max_length=128, verbose_name=_('E-Mail'), blank=True)
```
janibonnevier: Totally nitpicking, but I'd call this field `email`. Mainly because it is the name of the email field on the User model, but also because `mail` makes me think of traditional mail :)

BernhardPosselt: Right, however its already mail in the schema and throughout the entire code base xD

I think its fine as is :)

PR #153, opened 2016-07-23
Bump char fields to maximum of 256 chars since description/summary fi…

BernhardPosselt:
…elds prevented apps from being uploaded

@janibonnevier @adsworth @LukasReschke


janibonnevier:
:+1:

PR #154, opened 2016-07-25
Fix #152: Show nightlies on app detail page

### LINKED ISSUE #152: Nightlies on the app detail page
The app detail page lists the latest stable release but not the latest unstable release for a Nextcloud version. In addition if only nightlies exist there will be no link to the releases page at all.

There should be another column to the right of the current stable version which is colored in the same color like the nightly on the releases page if a nightly is available. That way you can always access the releases page, even when only nightlies exist for that Nextcloud version

janibonnevier:
#152

### DIFF
--- a/nextcloudappstore/core/templates/app/detail.html
+++ b/nextcloudappstore/core/templates/app/detail.html
@@ -93,11 +93,12 @@
             <section class="app-download">
                 <h4 id="downloads">{% trans "Downloads" %}</h4>
                 <table class="table">
-                    {% for platform_v, release in latest_releases_by_platform_v|sort_by_version:'desc' %}
-                        {% if release %}
+                    {% for platform_v, releases in latest_releases_by_platform_v|sort_by_version:'desc' %}
+                        {% if releases.stable or releases.nightly %}
                             <tr>
                                 <td>Nextcloud {{ platform_v }}</td>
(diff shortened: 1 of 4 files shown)


nextcloudappstore/core/templates/app/detail.html
```diff
@@ -93,11 +93,12 @@
...
                 <h4 id="downloads">{% trans "Downloads" %}</h4>
                 <table class="table">
-                    {% for platform_v, release in latest_releases_by_platform_v|sort_by_version:'desc' %}
-                        {% if release %}
+                    {% for platform_v, releases in latest_releases_by_platform_v|sort_by_version:'desc' %}
+                        {% if releases.stable or releases.nightly %}
                             <tr>
                                 <td>Nextcloud {{ platform_v }}</td>
-                                <td><a href="{{ release.download }}" rel="noopener noreferrer">Latest release ({{ release.version }})</a></td>
+                                <td>{% if releases.stable %}<a href="{{ releases.stable.download }}" rel="noopener noreferrer">Latest release ({{ releases.stable.version }})</a>{% endif %}</td>
```
BernhardPosselt: Maybe we should drop the preceding "Latest release" and "Nightly" and just use "x.x.x" and "x.x.x-nightly (unstable)" since it takes up a lot of screen space

BernhardPosselt:
Other than that looks good :)

PR #156, opened 2016-08-03
Migrate owncloud versions and add validation for more elements

BernhardPosselt:
Works in the following way:
If no \<nextcloud> tag exists the versions will be taken from \<owncloud>  and following will be done:
* 9.0 will be transformed to 9.0
* 9.1 will be transformed to 10
* 9.2 will be transformed to 11

I hope that's enough and we dont need to consider further exceptions

Furthermore:
* validation and docs for background-jobs, repair-steps and two-factor-providers elements (please review @ChristophWurst , I assumed two-factor-providers/two-factor-provider as element that contains the class name)
* validation will fail if a non allowed element is found (shipped, default_enable, standalone) or if deprecated elements are used (requiremin, requiremax, public, remote)

@nickvergessen @janibonnevier @adsworth @LukasReschke @MorrisJobke please review

### DIFF
--- a/nextcloudappstore/core/api/v1/release/info.xsd
+++ b/nextcloudappstore/core/api/v1/release/info.xsd
@@ -418,7 +418,7 @@
 
     <xs:complexType name="two-factor-providers">
         <xs:sequence>
-            <xs:element name="two-factor-provider" type="php-class" minOccurs="1"
+            <xs:element name="provider" type="php-class" minOccurs="1"
                         maxOccurs="unbounded"/>
         </xs:sequence>
     </xs:complexType>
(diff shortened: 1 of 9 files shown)


nextcloudappstore/core/api/v1/release/info.xsd
```diff
@@ -329,12 +329,12 @@
                         maxOccurs="unbounded"/>
             <xs:element name="lib" type="min-max-version" minOccurs="0"
                         maxOccurs="unbounded"/>
-            <xs:element name="owncloud" type="owncloud" minOccurs="1"
+            <xs:element name="nextcloud" type="nextcloud" minOccurs="1"
```
nickvergessen: min === 0 ?

BernhardPosselt: Minimum version is required (also in core afaik)

BernhardPosselt:
I've also added the following things:
* validation and docs for background-jobs and repair-steps elements
* validation will fail if a non allowed tag is found (shipped, default_enable, standalone)

BernhardPosselt:
More validation, see updated first post

BernhardPosselt:
:+1: from IRC

PR #165, opened 2016-08-11
Fix #116: Create app upload form

### LINKED ISSUE #116: Upload apps over the web interface
Uploading apps over the web interface should be trivial by providing a form which contains the download link, hash and nightly checkbox. The form should use the API through token authentication and should be available for logged in users under an **Upload app** link in the top navigation to the right of **Apps**

The token should be queried via POST from a non API route (important, we want CSRF protection since we are using session auth!), e.g.:

    POST /user/token

AFAIK POST is protected via CSRF by default (GET isnt), but you should verify that by sending a request without the token and check if it fails.

The form should be submitted using JS and error messages need to be rendered from the response (no translations I think, but thats fine)

janibonnevier:
#116 

JS probably needs some work. I copypasted `ObtainAuthToken` and modified it so that it returns the token based on `SessionAuthentication`, and not just the data sent for `BasicAuthentication`.

@BernhardPosselt


nextcloudappstore/core/api/v1/views.py
```diff
@@ -130,3 +133,30 @@ def get_object(self):
...
+
+class SessionObtainAuthToken(APIView):
+    """Modified version of rest_framework.authtoken.views.ObtainAuthToken.
+
+    Modified to return token based on SessionAuthentication, and not just data
+    sent for BasicAuthentication.
+    """
+
+    throttle_classes = (PostThrottle,)
+    permission_classes = (IsAuthenticated,)
```
BernhardPosselt: was that the only thing that was missing?

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+    'use strict';
+
+    function sendAppReleaseAPIRequest(download, checksum, nightly, token,
+            successCallback, failureCallback) {
+        let data = new Object();
+        data['download'] = download;
+        if (nightly) data['nightly'] = nightly;
+        if (checksum) data['checksum'] = checksum;
+
+        let xhr = new XMLHttpRequest();
```
BernhardPosselt: fetch api! see https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/static/assets/js/app-detail.js#L285

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+        }
+        xhr.send(JSON.stringify(data));
+    }
+
+    function clearMessages() {
+        let msgAreas = Array.from(document.querySelectorAll('[id$="-msg"]'));
+        msgAreas.forEach((el) => el.innerHTML = '');
+    }
+
+    function printMessages(response, statusCode) {
```
BernhardPosselt: http errors should be handled by the promise that the fetch api returns

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+        let result = 'failure alert-danger';
+        if (statusCode == 200 || statusCode == 201) {
+            result = 'success alert-success';
+        }
+
+        let prop;
+        for (prop in response) {
+            if (!response.hasOwnProperty(prop)) continue;
+            let msg;
+            if (typeof response[prop] == 'string') msg = response[prop];
```
BernhardPosselt: one line ifs bad!

BernhardPosselt: use var if you want function scope instead of let

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+            else if (response[prop] instanceof Array) msg = response[prop].join(' ');
+            document.getElementById(prop + '-msg').innerHTML =
+                '<div class="alert ' + result + '">' + msg + '</div>';
```
BernhardPosselt: XSS

nextcloudappstore/core/templates/nav.html
```diff
@@ -41,6 +41,11 @@
                 <li class="nav-link">
                     <a href="{% url 'home' %}">{% trans 'Apps' %}</a>
                 </li>
+                {% if request.user.is_authenticated %}
```
BernhardPosselt: is is_authenticated already a property? if so we should also convert the others to obtain django 1.10 compatability

nextcloudappstore/core/templates/app/upload.html
```diff
@@ -0,0 +1,29 @@
...
+
+{% block head %}
+    <script defer="defer" src="{% static 'assets/js/app-upload.js' %}"></script>
+{% endblock %}
+
+{% block apps %}
+    <h1>{% trans 'Upload App' %}</h1>
+    <section class="app-upload">
+        {% if request.user.is_authenticated %}
+        <form id="app-upload-form" method="post" action="{{ api_url }}">
```
BernhardPosselt: where do we get the variable from? maybe we should reverse the url with the url tag

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+                    download.value,
+                    checksum.value,
+                    nightly.checked,
+                    token,
+                    printSuccessMessage,
+                    printMessages);
+            });
+        }
+    }
+    xhr.send();
```
BernhardPosselt: doesnt this auto submit :D? 

I'd simply bind to the submit event https://developer.mozilla.org/de/docs/Web/Events/submit and then parse the fields, send the request and call preventDefault()? to prevent the default form submission

nextcloudappstore/core/static/assets/js/app-upload.js
BernhardPosselt: Either use a template engine or https://developer.mozilla.org/de/docs/Web/API/Document/createElement (which is probably better for such a short part). You could also render everything and use display: none + set the value if appropriate. Also take a look at https://developer.mozilla.org/de/docs/Web/HTML/Element/template

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    function printMessages(response, statusCode) {
+        let result = 'failure alert-danger';
+        if (statusCode == 200 || statusCode == 201) {
+            result = 'success alert-success';
+        }
+
+        let prop;
+        for (prop in response) {
```
BernhardPosselt: I suppose you want to iterate over a dict, correct? I'd pull in the ES7 shim using 

    bower install --save es7-shim

and use:

    for ([key, value] of Object.entries(response)) ...

You could also do:

    Object.keys(response).forEach((key) => response[key]

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+            else if (response[prop] instanceof Array) msg = response[prop].join(' ');
+            document.getElementById(prop + '-msg').innerHTML =
+                '<div class="alert ' + result + '">' + msg + '</div>';
+        }
+    }
+
+    function printSuccessMessage() {
+        let form = document.getElementById('app-upload-form');
+        form.parentNode.innerHTML =
+            '<div class="alert success alert-success">App release successfully uploaded.</div>';
```
BernhardPosselt: This can't be translated. I'd put this into the dom as either template or hidden element and use django's translation system

nextcloudappstore/core/api/v1/views.py
```diff
@@ -130,3 +133,30 @@ def get_object(self):
...
+    parser_classes = (parsers.FormParser,
+                      parsers.MultiPartParser,
+                      parsers.JSONParser,)
+    renderer_classes = (renderers.JSONRenderer,)
+    serializer_class = AuthTokenSerializer
+
+    def post(self, request, *args, **kwargs):
+        if request.user.is_authenticated():
```
BernhardPosselt: Can we already use **if request.user.is_authenticated** ?

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    function clearMessages() {
+        let msgAreas = Array.from(document.querySelectorAll('[id$="-msg"]'));
+        msgAreas.forEach((el) => el.innerHTML = '');
+    }
+
+    function printMessages(response, statusCode) {
+        let result = 'failure alert-danger';
+        if (statusCode == 200 || statusCode == 201) {
```
BernhardPosselt: Similarly this should just be in the sucess block of the promise

nextcloudappstore/core/api/v1/views.py
janibonnevier: I assume it would not return true if we couldn't? From REST API docs:
> If successfully authenticated, SessionAuthentication provides the following credentials.
>
> - request.user will be a Django User instance.
> - request.auth will be None.

http://www.django-rest-framework.org/api-guide/authentication/#sessionauthentication

I think `post()` is one of the last methods called in the class-
based view.

nextcloudappstore/core/api/v1/views.py
janibonnevier: The old `post()` only looked at the BasicAuthentication data when determining what token should be sent.

nextcloudappstore/core/templates/nav.html
janibonnevier: That's a method call because "template language" :)

nextcloudappstore/core/templates/app/upload.html
janibonnevier: Good catch. I tried some different things and forgot that there. The form itself does not need that attribute, as JS does the API call using the form values, but the form is not actually submitted. (is that bad?)

nextcloudappstore/core/api/v1/views.py
BernhardPosselt: Right, it used the serializer to log in (dafuq :D)

nextcloudappstore/core/api/v1/views.py
BernhardPosselt: I mean is_authenticated was converted to a property in Django 1.10

nextcloudappstore/core/templates/nav.html
BernhardPosselt: ok :)

Would you be so kind to grep for is_authenticated() :D? we still go that all over our template (see https://docs.djangoproject.com/en/1.10/releases/1.10/#user-is-auth-anon-deprecation)

nextcloudappstore/core/templates/app/upload.html
BernhardPosselt: Nope, but we could parse the URL from the form instead of hard coding it in our JS :)

May be more maintainable

BernhardPosselt: AFAIK you can prevent form submission with event.preventDefault() (where event is the first parameter for the submit event)

nextcloudappstore/core/static/assets/js/app-upload.js
BernhardPosselt: https://davidwalsh.name/fetch

nextcloudappstore/core/static/assets/js/app-upload.js
janibonnevier: > doesnt this auto submit :D? 

No, my dev server only gets one such request and that's when you click the button. But sure, your suggestion makes sense and additionally supports using the enter key.

nextcloudappstore/core/static/assets/js/app-upload.js
BernhardPosselt: Another possibility is to use a generator:
```js
function* entries(obj) {
    for (let key of Object.keys(obj)) {
        yield [key, obj[key]];
    }
}

for ([key, value] of entries({a: "b", b: "c"}) {
    console.log(key + " " + value);
}
```

nextcloudappstore/core/templates/nav.html
janibonnevier: Actually, we don't :)

```
$ grep -r "is_authenticated()" .
./venv/lib/python3.5/site-packages/rest_framework/permissions.py:        return request.user and request.user.is_authenticated()
./venv/lib/python3.5/site-packages/rest_framework/permissions.py:            request.user.is_authenticated()
./venv/lib/python3.5/site-packages/rest_framework/permissions.py:            (request.user.is_authenticated() or not self.authenticated_users_only) and
./venv/lib/python3.5/site-packages/rest_framework/throttling.py:        if request.user.is_authenticated():
./venv/lib/python3.5/site-packages/rest_framework/throttling.py:        if request.user.is_authenticated():
./venv/lib/python3.5/site-packages/rest_framework/throttling.py:        if request.user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/flatpages/templatetags/flatpages.py:            if not user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/flatpages/views.py:    if f.registration_required and not request.user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/auth/decorators.py:        lambda u: u.is_authenticated(),
./venv/lib/python3.5/site-packages/django/contrib/auth/mixins.py:        if not request.user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/auth/__init__.py:    if hasattr(user, 'is_authenticated') and not user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/auth/middleware.py:            if self.force_logout_if_no_header and request.user.is_authenticated():
./venv/lib/python3.5/site-packages/django/contrib/auth/middleware.py:        if request.user.is_authenticated():
./venv/lib/python3.5/site-packages/allauth/account/adapter.py:        assert request.user.is_authenticated()
./venv/lib/python3.5/site-packages/allauth/account/adapter.py:        if request.user.is_authenticated():
./venv/lib/python3.5/site-packages/allauth/account/views.py:        if request.user.is_authenticated() and \
./venv/lib/python3.5/site-packages/allauth/account/views.py:        if not self.request.user.is_authenticated():
./venv/lib/python3.5/site-packages/allauth/account/views.py:        if self.request.user.is_authenticated():
./venv/lib/python3.5/site-packages/allauth/account/tests.py:        self.assertTrue(user.is_authenticated())
./venv/lib/python3.5/site-packages/allauth/socialaccount/adapter.py:        assert request.user.is_authenticated()
./venv/lib/python3.5/site-packages/allauth/socialaccount/helpers.py:    if request.user.is_authenticated():
./nextcloudappstore/core/api/v1/views.py:        if request.user.is_authenticated():
./nextcloudappstore/core/permissions.py:        elif request.user and request.user.is_authenticated():
```

```
$ grep -r "is_authenticated" nextcloudappstore/core/templates/*
nextcloudappstore/core/templates/account/password_reset.html:    {% if user.is_authenticated %}
nextcloudappstore/core/templates/app/list.html:                {% if request.user.is_authenticated %}
nextcloudappstore/core/templates/app/upload.html:        {% if request.user.is_authenticated %}
nextcloudappstore/core/templates/nav.html:                {% if request.user.is_authenticated %}
nextcloudappstore/core/templates/nav.html:                {% if request.user.is_authenticated %}
```

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+        // User token request successful
+        form.addEventListener('submit', (event) => {
+            event.preventDefault();
+            apiRequestAppRelease(
+                url,
+                download.value,
+                checksum.value,
+                nightly.checked,
+                response.token)
+                .then((response) => {
```
BernhardPosselt: you can omit the braces if its a one liner, e.g.  .then(() => printSuccessMessage(), (response) -> printErrorMessage(response));

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+        var msgAreas = Array.from(document.querySelectorAll('[id$="-msg"]'));
+        msgAreas.forEach((el) => el.innerHTML = '');
+    }
+
+
+    function printErrorMessages(response) {
+        clearMessages();
+        Object.keys(response).forEach((key) => {
+            var msg;
+            if (typeof response[key] == 'string') {
```
BernhardPosselt: always try to use === and !== instead of == and != :)

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+                return json.then(Promise.reject.bind(Promise));
+            }
+        });
+    }
+
+
+    function apiRequestAppRelease(url, download, checksum, nightly, token) {
+        var data = {'download': download};
+        if (nightly) {
+            data['nightly'] = nightly;
```
BernhardPosselt: probably easier to just set it since its a boolean :)

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+                method: 'POST',
+                headers: new Headers({
+                    'Content-Type': 'application/json',
+                    'X-CSRFToken': csrf,
+                }),
+                credentials: 'include'
+            }
+        );
+
+        return fetch(request).then((response) => {
```
BernhardPosselt: What are you trying to do here? Why not simply 
    
    return fetch(request).then((response) => response.json());

if its not 200 its rejected by default afaik

nextcloudappstore/core/static/assets/js/app-upload.js
BernhardPosselt: you could even shorten it further:

    .then(printSuccessMessage, printErrorMessage);

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+    function apiRequestAppRelease(url, download, checksum, nightly, token) {
+        var data = {'download': download};
+        if (nightly) {
+            data['nightly'] = nightly;
+        }
+        if (checksum) {
+            data['checksum'] = checksum;
+        }
+
+        var request = new Request(
```
BernhardPosselt: let instead of var :)

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+    function clearMessages() {
+        var msgAreas = Array.from(document.querySelectorAll('[id$="-msg"]'));
+        msgAreas.forEach((el) => el.innerHTML = '');
+    }
+
+
+    function printErrorMessages(response) {
+        clearMessages();
+        Object.keys(response).forEach((key) => {
+            var msg;
```
BernhardPosselt: now that this is not function scoped anymore you can also use let xD

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,132 @@
...
+            if (typeof response[key] == 'string') {
+                msg = response[key];
+            } else if (response[key] instanceof Array) {
+                msg = response[key].join(' ');
+            }
+
+            var msgArea = document.getElementById(key + '-msg');
```
BernhardPosselt: let

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,129 @@
+(function (window) {
+    'use strict';
+
+
+    function apiRequestToken(csrf) {
+        var request = new Request(
```
BernhardPosselt: let

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,129 @@
...
+    let checksum = document.getElementById('checksum');
+    let nightly = document.getElementById('nightly');
+
+    // Get the auth token of the currently authenticated user and
+    // bind the app release API request to the form submit event.
+    apiRequestToken(csrf).then((response) => {
+        // User token request successful
+        form.addEventListener('submit', (event) => {
+            event.preventDefault();
+            apiRequestAppRelease(
```
BernhardPosselt: function name is very difficult to understand. Maybe change it to "uploadAppRelease"

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,129 @@
...
+        successMsg.removeAttribute('hidden');
+    }
+
+
+    // Get needed values and elements from form.
+    let form = document.getElementById('app-upload-form');
+    let url = form.action;
+    let csrf = Array.from(
+        document.getElementsByName('csrfmiddlewaretoken'))[0].value;
+    let download = document.getElementById('download');
```
BernhardPosselt: I'd move the parameters into the **form.addEventListener('submit', (event) => {** block, since they will only be read once. Think of uploading an app and pasting the wrong url. If you correct the URL in the input field, the url will still be the old one

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,129 @@
...
+        var successMsg = document.getElementById('form-success');
+        form.remove();
+        successMsg.removeAttribute('hidden');
+    }
+
+
+    // Get needed values and elements from form.
+    let form = document.getElementById('app-upload-form');
+    let url = form.action;
+    let csrf = Array.from(
```
BernhardPosselt: Array.from is not needed here I think. What about using document.querySelector('[name="csrfmiddlewaretoken"']:first-child).value

BernhardPosselt: ah :first-child is not even needed for querySelector,  document.querySelector('[name="csrfmiddlewaretoken"]').value should be enough

nextcloudappstore/core/static/assets/js/app-upload.js
```diff
@@ -0,0 +1,116 @@
...
+                method: 'POST',
+                headers: new Headers({
+                    'Content-Type': 'application/json',
+                    'Authorization': 'Token ' + token
+                }),
+                body: JSON.stringify(data)
+            }
+        );
+
+        return fetch(request).then((response) => {
```
BernhardPosselt: see line 17

BernhardPosselt:
Code looks good, now for the gui :)

* The form should be styled similar to the register form (error messages below, labels above input fields, same styles for inputs except the broken glyphicon placement xD): 

img

* When submitting the form the following should happen:
 * The button should get a loading spinner like here http://codepen.io/jmalatia/pen/HkmaA
 * All form fields and buttons need to be disabled so while its uploading people are unable to resubmit or change fields that might lead to inconsistencies

BernhardPosselt:
Upload feedback looks good, just pushed a fix to also report Http errors from requests. 

Only changed that I'd like to see is that the form is still visible after the upload, so uploading more than one version is faster/easier. Maybe we should also clear the form

BernhardPosselt:
Inputs should ofc only be cleared on success ;)

BernhardPosselt:
perfect!

PR #166, opened 2016-08-11
First javascript cleanup

BernhardPosselt:
Fixed some stuff and moved classes/reusable code into separate files

@janibonnevier


nextcloudappstore/core/templates/app/list.html
```diff
@@ -31,7 +31,7 @@ <h2 class="search-results-title">{% blocktrans with query=search_query %}Search
         {% endif %}
         <li role="presentation" class="filters">
             <form method="get" action="{{ request.path }}" name="filter-form" id="filter-form">
-                <input type="checkbox" class="auto-submit" id="featured-checkbox" name="featured" value="true" {% if request.GET.featured == 'true' %}checked{% endif %}>
+            <input type="checkbox" class="auto-submit" id="featured-checkbox" name="featured" value="true" {% if request.GET.featured == 'true' %}checked{% endif %}>
```
janibonnevier: Cosmetic nitpicking, but in my opinion this should be indented like it was :)

BernhardPosselt: ah right good catch. editor weirdness ;D

janibonnevier:
:+1:

PR #178, opened 2016-08-17
Tell devs about the tar.gz requirement

### LINKED ISSUE #176: Add information about supported archive format to app upload page
There is no indication on the upload page about the supported archive formats. One only finds out after having uploaded an archive.
It would be good to add a small note, probably directly in the field itself

@jancborchardt

oparoz:
Fixes #176


janibonnevier:
:+1:

PR #188, opened 2016-08-20
Add legal notice page

BernhardPosselt:
@karlitschek @LukasReschke whoever wants to work on this, please fill out the details in the legal.html in plain HTML right here https://github.com/nextcloud/appstore/compare/legal?expand=1#diff-299cdf8f91a371ff58e95096c3027898R10 ;)

### DIFF
--- a/nextcloudappstore/core/static/assets/css/style.css
+++ b/nextcloudappstore/core/static/assets/css/style.css
@@ -413,3 +413,7 @@ form .text-danger {
     position: relative;
     top: -1px;
 }
+
+address {
+    margin-top: 15px;
+}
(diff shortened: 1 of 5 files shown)


karlitschek:
Easiest is probably to only link to this page: https://nextcloud.com/impressum/  This this possible?

BernhardPosselt:
@karlitschek yes, however I think it looks prettier if its embedded in the page itself. The contents probably won't change so I think its fine if we hard code it.

I've copied the stuff over from the nextcloud.com page and added some microformat markup, please review.

karlitschek:
perfect :-)

PR #189, opened 2016-08-21
API token regeneration

janibonnevier:
- New API route at `/api/v1/token/new/`, which generates a new API token for the authenticated user regardless of whether the user already has a token.
- Account page with form that sends requests to said API route.

### DIFF
--- a/nextcloudappstore/core/static/assets/js/account-api-token.js
+++ b/nextcloudappstore/core/static/assets/js/account-api-token.js
@@ -1,0 +1,96 @@
+(function (global) {
+    'use strict';
+
+    function regenAuthToken(url, api_token) {
+        let request = new Request(
+            url,
+            {
+                method: 'POST',
+                headers: new Headers({
+                    'Content-Type': 'application/json',
+                    'Authorization': 'Token ' + api_token
+                }),
+            }
+        );
+        return fetch(request).then(global.convertResponse);
+    };
+
+
+    function showElement(element, boolean) {
+        if (boolean) {
+            element.removeAttribute('hidden');
+        } else {
+            element.setAttribute('hidden', 'true');
+        }
+    }
+
+
+    function showTokenFetchFailureMessage(boolean) {
+        let msg = document.getElementById('token-failure');
+        let paragraphs = Array.from(document.querySelectorAll('#tokenSection p'));
+        showElement(msg, boolean);
+        paragraphs.forEach((p) => showElement(p, !boolean));
+    }
+
+
+    function showTokenRegenSuccessMessage(boolean) {
+        let msg = document.getElementById('regen-success');
+        showElement(msg, boolean);
+    }
+
+
+    function showTokenRegenFailureMessage(boolean) {
+        let msg = document.getElementById('regen-failure');
+        showElement(msg, boolean);
+    }
+
+
+    function updateTokenDisplay(token) {
+        let tokenEl = document.getElementById('token');
+        tokenEl.innerHTML = token;
+    }
+
+
+    function updateToken(csrf) {
+        global.fetchAPIToken(csrf).then(
+            (response) => {
+                updateTokenDisplay(response.token);
+                showTokenFetchFailureMessage(false);
+            },
+            showTokenFetchFailureMessage(true)
+        );
+    }
+
+
+    function onTokenRegenSuccess(response) {
+        showTokenRegenSuccessMessage(true);
+        showTokenFetchFailureMessage(false);
+        updateTokenDisplay(response.token);
+    }
+
+
+    function onTokenRegenFailure() {
+        showTokenRegenFailureMessage(true);
+    }
+
+
+    let form = document.getElementById('api-token-regen-form');
+    let csrfEl = document.getElementsByName('csrfmiddlewaretoken')[0];
+    let tokenEl = document.getElementById('token');
+    let confirmText = document.getElementById('regen-confirm-text').innerHTML;
+
+    updateToken(csrfEl.value);
+
+    form.addEventListener('submit', (event) => {
+        event.preventDefault();
+        if (confirm(confirmText)) {
+            showTokenRegenSuccessMessage(false);
+            showTokenRegenFailureMessage(false);
+            regenAuthToken(form.action, tokenEl.innerHTML).then(
+                (response) => onTokenRegenSuccess(response),
+                onTokenRegenFailure
+            );
+        }
+    });
+
+}(this));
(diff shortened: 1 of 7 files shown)


nextcloudappstore/core/static/assets/js/account-api-token.js
```diff
@@ -0,0 +1,96 @@
...
+    function showElement(element, boolean) {
+        if (boolean) {
+            element.removeAttribute('hidden');
+        } else {
+            element.setAttribute('hidden', 'true');
+        }
+    }
+
+
+    function showTokenFetchFailureMessage(boolean) {
```
BernhardPosselt: never use types as variable names. This should be something like showFailureMessage or isFailure

BernhardPosselt: the general question is if the parameter would need to be set at all. I don't fully understand what its used for

BernhardPosselt:
Looks good but the button is styled differently than everywhere else. We should probably also mention what the button is used to and what the effects of changing the token is:

* Token can be used for authenticating you when calling app store APIs
* Regenerating the token means that all open pages should be refreshed since we fetch the token on page load (currently limited to app upload and token regen page)

The second notice should not exist. Instead we should refetch the token before we make a request that requires a token, e.g.:
* User clicks upload app
* We fetch the token
* We use the token to upload the app

nextcloudappstore/core/static/assets/js/account-api-token.js
janibonnevier: I use the `hidden` attribute to hide messages and other elements that should not be displayed. When I want to show them, I remove the attribute.

janibonnevier: The parameter determines whether the element will be shown or hidden.

BernhardPosselt: When calling showTokenFetchFailureMessage, don't i always want to show it?

janibonnevier: True, I just followed the pattern for the other messages, which we sometimes do want to hide. Should I fix?

BernhardPosselt: yes

boolean flags are in general hard to read for functions because its not obvious what they are used for. Even better if you can avoid them ;)

BernhardPosselt: If functions are always called with the same value and you can call them in a wrong way by passing a different parameter you are actually making the API harder to use and therefore worse ;)

BernhardPosselt: Thats also the reason why you use enumerations instead of simple integers :) Poka yoke!

janibonnevier:
The reasoning behind using the `btn-default` class was to make the button less prominent. As I've understood, the primary purpose of the page is to display the token, while the regeneration is a secondary feature. I thought that the bright `btn-primary` stole all the attention.

BernhardPosselt:
@janibonnevier fully agree that its a very simple function and different from the other ones. However the way it looks now does not fit the overall design (every account page has a blue submit button at the bottom). Since it's the only thing on that page I don't see what else should be highlighted, I mean you make it less prominent in order to make something more prominent right?

janibonnevier:
Yes, I intended to make the token itself more prominent. But sure, I see your point. Fixing :)

nextcloudappstore/core/static/assets/js/account-api-token.js
```diff
@@ -0,0 +1,92 @@
...
+
+    function showTokenFetchFailureMessage() {
+        let msg = document.getElementById('token-failure');
+        let elementsToHide = Array.from(document.querySelectorAll('#tokenSection .hide-on-token-failure'));
+        showElement(msg, true);
+        elementsToHide.forEach((el) => showElement(el, false));
+    }
+
+
+    function showTokenRegenSuccessMessage(show) {
```
BernhardPosselt: Will this function ever be called with false? Same for the failure message

BernhardPosselt: ah, found it :)

nextcloudappstore/core/templates/user/api-token.html
```diff
@@ -2,10 +2,32 @@
...
 
 {% block account-content %}
     <h1>{% trans "API Token" %}</h1>
-    <section>
+    <section id="tokenSection">
+        <div id="regen-success" class="alert alert-success" hidden>{% trans 'New API token generated.' %}</div>
+        <div id="regen-failure" class="alert alert-danger" hidden>{% trans 'Failed to regenerate API token.' %}</div>
+        <div id="token-failure" class="alert alert-danger" hidden>{% trans 'Failed to fetch token.' %}</div>
+        <span id="regen-confirm-text" hidden>{% trans 'Do you want to regenerate your API token?' %}</span>
+        {% blocktrans %}
```
BernhardPosselt: I'd prefer trans since translators need to handle HTML otherwise. If you insist on blocktrans, at least move it inside the p tag :)

BernhardPosselt:
Looks good! Can be merged if the blocktrans thing is fixed

BernhardPosselt:
Another thing: the success message currently is always shown above everything else. The token however shows the success message below the heading. To reproduce: Change your password/name, then regenerate the token. Watch the position of the green message.

Place the message wherever you like but it should be consistent ;)

BernhardPosselt:
IMHO you could also drop the success messages since the token is changed anyways. In that case I'd also drop the confirm alert so the user sees that the token is changin :)

janibonnevier:
I know that the placement of the messages is inconsistent. The difference is where the message originates from, a Django view or JS. Either the message divs need to be moved to `base.html` (the most basic one) or the templates need to be rearranged so that the message section which now is in `base.html` is repeated in multiple templates. A third option would perhaps be to move the divs to the real message section with JS.

I do think that there should be a success message upon regeneration. I found that the token change is quite easy to miss, and I'd rather have too much feedback than too little.

Fixing the blocktrans.

nextcloudappstore/core/templates/user/api-token.html
janibonnevier: Fixed. But yeah, I think i insist on blocktrans in order to keep the line length manageable :)

BernhardPosselt:
The notification stuff can be moved to another ticket

BernhardPosselt:
:+1:

PR #191, opened 2016-08-21
Password change form

janibonnevier:


nextcloudappstore/core/templates/user/form-fields.html
```diff
@@ -0,0 +1,31 @@
...
+                <p class="text-danger">{{ error }}</p>
+            {% endfor %}
+        {% endif %}
+
+        {% if field.help_text %}
+            <p class="help-block">{{ field.help_text }}</p>
+        {% endif %}
+    </div>
+{% endfor %}
+{% if redirect_field_value %}
```
BernhardPosselt: probably not needed

BernhardPosselt: or is it?

BernhardPosselt:
Works and looks good for non social account passwords.

The menu entry should be hidden/disabled for social accounts however

BernhardPosselt:
I don't know what's the best way to tell if a user is logged in via github but i found
    
    request.user.has_usable_password()

which returns False if the user is logged in via GitHub and True if he is registered on our site

nextcloudappstore/core/templates/user/form-fields.html
janibonnevier: @adsworth might know.

adsworth: Hi,

I guess we don't need it there.

I think it might makes sense to use that include in the our other forms as well though. If I remember correctly all we do in our other forms is loop over the form and style via css.

janibonnevier:
Is it sufficent to hide the menu item and form from the templates, or should we truly disable the password changing functionality in the view?

BernhardPosselt:
Hiding is enough

BernhardPosselt:
:+1:

PR #192, opened 2016-08-23
Make app list entry all of the same height

BernhardPosselt:
@janibonnevier @LukasReschke 

bring back the symetry. Tested with the currently uploaded apps. Should work fine except when someone decides to do an extremely wide screen screenshot.


BernhardPosselt:
We should optimize for 16:9 screenshots and let app devs choose how they want to handle it imo. Alternative is ofc to download all screenshots when you're uploading an app and validate the ratio.

nextcloudappstore/core/templates/app/list.html
```diff
@@ -47,20 +47,19 @@ <h2 class="search-results-title">{% blocktrans with query=search_query %}Search
     </ul>
 </div>
 {% for app in object_list %}
-        <div class="col-lg-6 app-list-container">
+        <div class="col-md-6 app-list-container">
             <div class="well">
+                <div class="app-list-screenshot">
                 {% if app.screenshots.all %}
```
janibonnevier: Indentation?

PR #199, opened 2016-08-25
Fix #197: Consistent message placement

### LINKED ISSUE #197: Consistent message placement
Django views place messages in `div#body > div.messages`. On some pages (such as API token regeneration and app upload) we show messages that appear after page load, specifically feedback following AJAX requests. For simplicity they are currently shown in places other than the real message div. Something should be done about this. Perhaps it could be solved with template blocks.

janibonnevier:
#197 

- Create template block for messages
- Fix message placement on token regen page

I did not touch the success message on the app upload page because I think it looks way better as it is. That page is very different from the rest anyway.


BernhardPosselt:
👍

PR #201, opened 2016-08-26
Delete user account form

### LINKED ISSUE #83: Delete user account
A user should be able to delete his account by hitting and confirming it on the download page. The confirmation should not be able to trigger by accident, Github's delete repo ui is a good example.

Before deleting his account, a user will be warned that all his comments and apps will be deleted.

BernhardPosselt:
Fix #83 

@janibonnevier

### DIFF
--- a/nextcloudappstore/core/templates/user/delete-account.html
+++ b/nextcloudappstore/core/templates/user/delete-account.html
@@ -1,6 +1,6 @@
 {% extends "user/base.html" %}
 {% load i18n %}
-{% load account socialaccount %}
+{% load account socialaccount css_class %}
 
 {% block head-title %}{% trans "Delete account" %} - {% endblock %}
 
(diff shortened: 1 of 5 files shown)


nextcloudappstore/core/templates/user/delete-account.html
```diff
@@ -0,0 +1,35 @@
...
+
+{% block head-title %}{% trans "Delete account" %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "Delete account" %}</h1>
+    <section>
+        <form method="POST" action="{% url 'account-deletion' %}">
+            {% csrf_token %}
+            <p class="alert alert-danger">{% trans 'Deleting your account will delete all your apps and releases as well. To delete your account, insert your e-mail address into the input field and click the "Delete account" button:' %}</p>
+            {% for field in form %}
```
janibonnevier: I'm on mobile so it's cumbersome for me to check, but is it possible to include `user/form-fields.html` to replace most of the contents of this `<form>`?

BernhardPosselt: Oh, nice :)

nextcloudappstore/core/templates/user/delete-account.html
```diff
@@ -0,0 +1,16 @@
...
+
+{% block head-title %}{% trans "Delete account" %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "Delete account" %}</h1>
+    <section>
+        <p class="alert alert-danger">{% trans 'Deleting your account will delete all your apps and releases as well. To delete your account, insert your e-mail address into the input field and click the "Delete account" button:' %}</p>
+        <form method="POST" action="{% url 'account-deletion' %}">
+            {% include 'user/form-fields.html' %}
+            <p><input type="submit" class="btn btn-primary btn-block" value="{% trans 'Delete account' %}"></p>
```
janibonnevier: Just a design suggestion: Instead of `btn-primary` we could maybe use `btn-danger` here, which makes the button red. In that case `.btn-danger` should be added to [this rule in theme.css](https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/static/assets/css/theme.css#L65).

nextcloudappstore/core/user/views.py
```diff
@@ -3,9 +3,30 @@
...
+    def get_context_data(self, **kwargs):
+        context = super().get_context_data(**kwargs)
+        context['form'] = DeleteAccountForm()
+        context['acc_page'] = 'delete-account'
+        return context
+
+    def post(self, request, *args, **kwargs):
+        form = DeleteAccountForm(request.POST, user=request.user)
+        if form.is_valid():
+            self.request.user.delete()
```
janibonnevier: I don't think this `self` is needed here.

janibonnevier:
Do what you will about my two latest comments, otherwise :+1:  :)

BernhardPosselt:
Good suggestions, fixed both

PR #204, opened 2016-08-27
Fix #202

### LINKED ISSUE #202: Category for 2FA plugins
We have several plugins for our new **Pluggable 2-Factor Authentification** ... In the future there will be some more. 

Examples:

- **SMS** (https://github.com/ChristophWurst/twofactor_sms)
- **TOTP** (https://github.com/ChristophWurst/twofactor_totp)
- **U2F** (https://github.com/ChristophWurst/twofactor_u2f)

I think we should organize them in a seperate group in our appstore. What do you think?

cc @BernhardPosselt @ChristophWurst

BernhardPosselt:
@janibonnevier @mar1u5

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -207,7 +207,7 @@ category
 
     * old categories are migrated:
 
-       * **tool**, **game**, **other** will be converted to **tools**
+       * **tool**, **game** and **other** will be converted to **tools**
        * **productivity** will be converted to **organization**
 
     * can occur more than once with different categories
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -192,7 +192,8 @@ documentation/developer
 category
     * optional
     * if not provided the category **tools** will be used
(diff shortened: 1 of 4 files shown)


docs/developer.rst
```diff
@@ -192,7 +192,8 @@ documentation/developer
 category
     * optional
     * if not provided the category **tools** will be used
-    * must contain one of the following values: **customization**, **files**, **integration**, **monitoring**, **multimedia**, **office**, **organization**, **social**, **tools**
+    * must contain one of the following values: **auth**, **customization**, **files**, **integration**, **monitoring**, **multimedia**, **office**, **organization**, **social**, **tools**
+    * old categories are migrated (however: duplicate result categories are invalid!): **tool**, **game**, **other** -> **tools**, **productivity** -> **organization**
```
janibonnevier: It seems a bit unclear to me what categories are migrated to what. I would make one sub list item for each "->". Also, I have not really grasped what it is that is invalid :)

PR #208, opened 2016-08-28
Style social sign up

### LINKED ISSUE #135: Style social signup
Currently social signup looks like this

style

This does not fit the overall style on the website. To change that, override the signup template in templates/socialaccount/signup.html by copying the template from https://github.com/pennersr/django-allauth/blob/master/allauth/templates/socialaccount/signup.html and making adjustments similar to https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/templates/account/signup.html

BernhardPosselt:
Fix #135 

@LukasReschke @janibonnevier 

I've moved the form-fields include to the top because its very re-usable, then copied the affected templates over from allauth and added our classes:

signup


janibonnevier:
Haven't tried on my setup but seems good :)

PR #209, opened 2016-08-28
Rating foundation

BernhardPosselt:
Adds the backend basics without any controller/view hooks. This will allow the current API to expose a default rating of 0.5 which can be adjusted later on.

Another PR will follow which will add the controllers and gui interaction. 

The basic idea is to calculate a new app rating after every new rating (because its very expensive) for both recent (in the last 90 days) and overall.


BernhardPosselt:
@janibonnevier @LukasReschke @oparoz

nextcloudappstore/core/models.py
```diff
@@ -196,6 +202,58 @@ def _latest(self, releases):
             return None
 
 
+class AppRating(TranslatableModel):
+    app = ForeignKey('App', related_name='app', verbose_name=_('App'),
```
janibonnevier: I believe `related_name` should be something like `app_rating`. Same on line 209.

janibonnevier: Or rather, "ratings".

nextcloudappstore/core/rating.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def compute_rating(ratings: List[float], threshold: int = 10) -> float:
+    """
+    Turns a list of ratings into a score, 0 being the lowest and 1 being the
+    highest. The idea is that everyone can use his own custom scale, e.g.
+    a 5 star rating would multiply the value by 5. Also a too low number of
+    ratings will default to a constant value.
+    :param ratings: list of floats from 0 to 1
+    :param threshold: if the amount of ratings is lower than this number
```
janibonnevier: "lower than or equal to" according to the if-statement.

nextcloudappstore/core/models.py
```diff
@@ -196,6 +202,58 @@ def _latest(self, releases):
...
+        super().save(*args, **kwargs)
+        # update rating on the app
+        app = self.app
+        day_range = settings.RATING_RECENT_DAY_RANGE
+        threshold = settings.RATING_THRESHOLD
+        app.rating_recent = self._compute_app_rating(app, day_range, threshold)
+        app.rating_overall = self._compute_app_rating(app, threshold=threshold)
+        app.save()
+
+    def _compute_app_rating(self, app: App, days: int = -1,
```
janibonnevier: Isn't `app` always the same as `self.app`? In other words: how come this method is basically static? And if we want it to be static, does it belong in a `Model` class?

janibonnevier:
Apart from my nitpick on IRC, :+1:

PR #213, opened 2016-08-28
Adjust string lengths in schema

BernhardPosselt:
Turns out string lengths were not adjusted in the schema when the models were changed. Every VARCHAR has a length of 256 chars now, this should now also be reflected in the schema.

@ChristophWurst @LukasReschke @janibonnevier 

That makes the mail app almost valid :D


ChristophWurst:
LGTM

Issue #193, opened 2016-08-25
Fetch API token on app upload


janibonnevier:
Currently on the app upload page the API token is fetched on page load. It should instead be fetched if and when the form is submitted. This would allow the form to function properly even if the token is regenerated between the page load and the submission.

Token fetching and form submission is done in https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/static/assets/js/app-upload.js using functions defined in https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/static/assets/js/utils.js.

PR #217, opened 2016-08-30
Fetch API token on form submit for app upload

clone1612:
Fix for #193 by switching the location of the "token fetch" and the binding of the form submit event. 

If anything is wrong, feel free to say so. I have the habit of seeing a problem as trivial and fabricating a solution that is not adequate :)

Code was tested and comments were added where needed.


BernhardPosselt:
:+1: tested locally and it works like described in the issue

Great job @clone1612 !

PR #219, opened 2016-09-02
Fix screenshot height for images with small height

juliushaertl:
This fixes images that get stretched if their height is to small.

Before (see keeweb screenshot)
2016-09-02-111229_841x348_scrot

After
2016-09-02-111210_834x345_scrot


janibonnevier:
:+1:

BernhardPosselt:
Very good idea!

BernhardPosselt:
Scale images correctly but keep the container height

BernhardPosselt:
BTW maybe at some point we will also need to vertically center images

PR #220, opened 2016-09-03
Let users rate app

### LINKED ISSUE #212: Let users rate an app
On the app detail page, if logged in, the user should see a filled out form with his current rating or an empty form with the default rating "Ok" above the list of comments. Implement it like a normal form, meaning: hitting submit should reload the page. There are three possible ratings for the user to choose from:
* Bad (0.0)
* Ok (0.5)
* Good (1.0)

Ideally we also have a visual way to represent those (thumbs or smilies?)

In the Textfield the user should be able to write a rating comment in Markdown.

Another thing: we should make it clear to users to not post questions, feature requests or bug reports by explaining that the app maintainer will not be notified by rating comments.

Additionally we should think about how to prevent users from posting bug reports or questions in the rating comments. Maybe hide the form and add 4 buttons, reading: "Report problem", "Request feature", "Ask question", "Rate app". The first two should redirect to the bug tracker (hide buttons if absent) third one should redirect to the discussion in the forum (hide buttons if absent). The fourth one will display the rating form.

BernhardPosselt:
Fix #212

Things are a bit ugly because model form parsing has to be done by hand (language for the comment should be the current user's language)

I'm not 100% happy with how it turned out, let's see :)

@janibonnevier


BernhardPosselt:
ratings

nextcloudappstore/core/views.py
```diff
@@ -29,8 +32,33 @@ class AppDetailView(DetailView):
...
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
+        try:
+            app_rating = AppRating.objects.get(user=self.request.user,
+                                               app=context['app'])
+            context['rating_form'] = AppRatingForm(initial={
+                'rating': app_rating.rating,
+                'comment': app_rating.comment
+            })
+            context['user_rated_app'] = True
```
janibonnevier: "user_rated_app" sounds like noun. I'd name this something like "user_has_rated" or "user_has_rated_app", although "app" might be a bit redundant since we are in the app detail view.

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -110,5 +103,33 @@ <h4 id="downloads">{% trans "Downloads" %}</h4>
             </section>
         </div>
     </div>
+    {% if object.issue_tracker or object.discussion or request.user.is_authenticated %}
+        <div class="row app-comments">
+            <ul class="col-md-12">
+                <h4 class="section-heading" id="comments">{% trans 'Comments' %}</h4>
```
janibonnevier: Feels like this section is broader than just "Comments". I can't come up with a better alternative, though.

janibonnevier: "Discussion"?

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -110,5 +103,33 @@ <h4 id="downloads">{% trans "Downloads" %}</h4>
...
+                {% if object.issue_tracker %}
+                    <li><a target="_blank" rel="noreferrer noopener" href="{{ object.issue_tracker }}" class="btn btn-secondary">{% trans 'report problem' %}</a></li>
```
janibonnevier: I think these button texts should start with a capital letter like all the previous ones around the site. Or we can somehow rationalize the inconsistency.

janibonnevier:
:+1:

PR #239, opened 2016-09-05
Require first and last name when signing up

BernhardPosselt:
So users are not automatically marked as Anonymous

Tested with both github and local sign up

@janibonnevier


janibonnevier:
:+1:

janibonnevier:
:+1: :D

PR #264, opened 2016-09-09
Also persist raw versions to make it easier for nextcloud to implemen…

BernhardPosselt:
…t it

@LukasReschke @janibonnevier


janibonnevier:
:+1:

PR #265, opened 2016-09-09
First data model changes for signatures

### LINKED ISSUE #256: API: Expose app certificate
The certificate should be exposed in the api. Serializers and docs need to be adjusted.

BernhardPosselt:
Nothing is validated yet, only the data model, docs and upload form are adjusted

Fix #256 

@janibonnevier @LukasReschke


janibonnevier:
Nice work :) :+1:

PR #270, opened 2016-09-09
Fix #267: Delete all releases of an app upon certificate change

### LINKED ISSUE #267: API: changing the certificate for an app should delete all of its releases
Since they can not be validated anymore

janibonnevier:
#267

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -203,6 +203,17 @@ class App(TranslatableModel):
         except ValueError:
             return None
 
+    def save(self, *args, **kwargs):
+        # If the certificate has changed, delete all releases.
+        try:
+            if self.pk is not None:
+                orig = App.objects.get(pk=self.pk)
+                if self.certificate != orig.certificate:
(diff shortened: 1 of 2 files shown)


nextcloudappstore/core/models.py
```diff
@@ -203,6 +203,17 @@ def _latest(self, releases):
...
             return None
 
+    def save(self, *args, **kwargs):
+        # If the certificate has changed, delete all releases.
+        try:
+            if self.pk is not None:
+                orig = App.objects.get(pk=self.pk)
+                if self.certificate != orig.certificate:
+                    self.releases.all().delete()
+        except self.DoesNotExist:
```
janibonnevier: I don't understand where this exception class is.

BernhardPosselt: App.DoesNotExist

BernhardPosselt:
:+1:

nextcloudappstore/core/models.py
janibonnevier: Apparently...

PR #274, opened 2016-09-10
Add pyopenssl

BernhardPosselt:


BernhardPosselt:
@LukasReschke tested it on an Ubuntu box and it installs :D

You need another package which was added to the docs

PR #308, opened 2016-09-21
Split out typing

BernhardPosselt:
Ready to review


BernhardPosselt:
@LukasReschke like discussed this script should? get rid of the venv and install everything from scratch to make updates fail less often

Issue #131, opened 2016-07-14
Allow transfer of app ownership


janibonnevier:
I believe it should be possible to transfer ownership of an app from one user to another. There are several scenarios where this would be relevant, e.g. if an app is sold or handed over to someone else to manage. A developer might also want to move an app between two of his own accounts, say for example that he has one personal account and one for his business. 

Obviously such an action should only be initiated by the current owner of the app, but the potential new owner must also be able to accept or decline. 

Another thing to consider is whether the users of an app would like to know about any changes in ownership.

BernhardPosselt:
:+1:

andyboeh:
I would love to see this fixed, as I need to transfer the ownership of an app to a new maintainer. Until then, what's the best way to transfer app ownership?

janibonnevier:
Sorry, I've been too busy to work on this :( 

Right now you need an app store administrator to change the owner manually for you. Unsure whether they will see this here, but otherwise you should be able to find them on IRC (Freenode #nextcloud-dev).

BernhardPosselt:
@andyboeh just create a ticket. We probably will verify it using a signature over some value using the app's certificate. More on that in a separate ticket.

PR #310, opened 2016-09-23
Model for transfer of app ownership

janibonnevier:
#131


nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+
+        AppOwnershipTransfer.objects.create(app=app, new_owner=user)
+
+    An instance of AppOwnershipTransfer is deleted when the transfer it
+    represents is committed.
+    """
+
+    app = OneToOneField(
+        App, on_delete=CASCADE, related_name='ownership_transfer')
+    old_owner = ForeignKey(
```
BernhardPosselt: would probably name this field **from**

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+        User,
+        related_name='app_ownership_transfers_outgoing',
+        on_delete=CASCADE,
+        null=False)
+    new_owner = ForeignKey(
```
BernhardPosselt: and this field **to**

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+    following statement:
+
+        AppOwnershipTransfer.objects.create(app=app, new_owner=user)
+
+    An instance of AppOwnershipTransfer is deleted when the transfer it
+    represents is committed.
+    """
+
+    app = OneToOneField(
+        App, on_delete=CASCADE, related_name='ownership_transfer')
```
BernhardPosselt: Models should be in wrapped in ', e.g. 'App' rather than App

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+        User,
+        related_name='app_ownership_transfers_outgoing',
+        on_delete=CASCADE,
+        null=False)
+    new_owner = ForeignKey(
+        User,
+        related_name='app_ownership_transfers_incoming',
+        on_delete=CASCADE,
+        null=False)
+    is_accepted_old_owner = BooleanField(default=False)
```
BernhardPosselt: Probably irrelevant to store if both users accepted it. If a request is created we will render it. If its approved we will delete the entry and change the owner on the app

BernhardPosselt: You do that further todwn btw

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+
+    def save(self, force_insert=False, force_update=False, using=None,
+             update_fields=None):
+        if not self.id:  # a.k.a. "if object is being created"
+            self.old_owner = self.app.owner
+            if self.old_owner is self.new_owner:
+                raise RuntimeError(
+                    'Could not initiate transfer of app ownership. '
+                    + 'The proposed new owner already owns the app.')
+        return super().save()
```
BernhardPosselt: arguments are not passed to super

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,62 @@ class Meta:
...
+    represents is committed.
+    """
+
+    app = OneToOneField(
+        App, on_delete=CASCADE, related_name='ownership_transfer')
+    old_owner = ForeignKey(
+        User,
+        related_name='app_ownership_transfers_outgoing',
+        on_delete=CASCADE,
+        null=False)
```
BernhardPosselt: https://docs.djangoproject.com/en/1.10/ref/models/fields/#null default is already False

nextcloudappstore/core/models.py
BernhardPosselt: and use settings.AUTH_USER_MODEL as user foreign key

janibonnevier: `from` is a Python keyword. `from_user`?

nextcloudappstore/core/models.py
janibonnevier: I put it there for flexibility. But maybe we don't need that :)

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,54 @@ class Meta:
...
+        'App', on_delete=CASCADE, related_name='ownership_transfer')
+    from_user = ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='app_ownership_transfers_outgoing',
+        on_delete=CASCADE)
+    to_user = ForeignKey(
+        settings.AUTH_USER_MODEL,
+        related_name='app_ownership_transfers_incoming',
+        on_delete=CASCADE)
+    is_accepted = BooleanField(default=False)
```
BernhardPosselt: we can probably ditch that field

nextcloudappstore/core/models.py
```diff
@@ -512,3 +512,54 @@ class Meta:
...
+        self.app.owner = self.to_user
+        self.app.save()
+        self.delete()
+
+    def save(self, *args, **kwargs):
+        if not self.id:  # a.k.a. "if object is being created"
+            self.from_user = self.app.owner
+            if self.from_user is self.to_user:
+                raise RuntimeError(
+                    'Could not initiate transfer of app ownership. '
```
BernhardPosselt: the + is not needed inside () afaik to concat strings

nextcloudappstore/core/tests/test_app_ownership_transfer.py
```diff
@@ -0,0 +1,23 @@
+from django.contrib.auth.models import User
+from django.test import TestCase
+from nextcloudappstore.core.models import App, AppOwnershipTransfer
+
+
+class AppOwnershipTransferTest(TestCase):
+    def setUp(self):
+        self.user1 = User.objects.create_user(username='user1')
```
BernhardPosselt: AFAIK things run in setUp are not included in the transaction and you need to delete them again in the tearDown method. I'd probably move them to a separate method and call that one in each relevant block. Alternative is to delete them in a tearDown method

PR #326, opened 2016-09-29
Add very simple ocs api

BernhardPosselt:
@LukasReschke


BernhardPosselt:
Closed because we cant generate integer ocs ids and Nextcloud has completely useless int casting all over the place

BernhardPosselt:
@janibonnevier please also review ;)

PR #328, opened 2016-10-02
Parse and expose changelog in API

### LINKED ISSUE #86: App changelog integration
Some apps provide a CHANGELOG.md inside the app folder which follows a specific format. It would be awesome if we could parse the markdown and add it to the appropriate release overview.

Things to keep in mind:
* The file needs to be streamed when extracted from the archive to prevent zip bombs
* The file must be checked for hard or symlinks which are not allowed (should throw)
* Take a look at the info.xml handling

BernhardPosselt:
Fix #86 

@janibonnevier @LukasReschke


BernhardPosselt:
Also cc @jhass @georgehrke @raghunayyar @Henni @ChristophWurst @MorrisJobke 

Is the proposed Changelog format ok for you?

ChristophWurst:
The mail app uses a structure inspired by http://keepachangelog.com/en/0.3.0/: https://github.com/nextcloud/mail/blob/master/CHANGELOG.md

BernhardPosselt:
@ChristophWurst thanks for the tip, looks like a sane format and it's always better to choose a standard for this. Updated the parser to use the linked format.

BernhardPosselt:
BTW, the standard also answered my question for "what to do with nightlies". Nightly changelogs are now parsed from the **## [Unreleased]** section

PR #331, opened 2016-10-04
Import OCS ids if present in info.xml

BernhardPosselt:
This also imports OCS ids from the info.xml and generates integer ids for categories

Makes https://github.com/nextcloud/server/pull/1598 obsolete, Nextcloud 9 and 10 should be able to switch to the new store without modifications.

Apps have to be reuploaded to add the OCS id. This also ensures that apps can be updated if the same id is used


BernhardPosselt:
@nickvergessen @MorrisJobke @janibonnevier @LukasReschke

BernhardPosselt:
@georgehrke

PR #333, opened 2016-10-05
Make changelogs translatable

BernhardPosselt:
Only english changelogs are probably relevant for now but adding it now will prevent api breakage and will make it future proof :)


BernhardPosselt:
@LukasReschke @janibonnevier

PR #337, opened 2016-10-06
Style admin

BernhardPosselt:
@janibonnevier @LukasReschke


nextcloudappstore/core/static/assets/css/admin/admin.css
```diff
@@ -0,0 +1,107 @@
+@font-face {
+    font-family: 'Glyphicons Halflings';
+    src: url('../../../vendor/bootstrap/dist/fonts/glyphicons-halflings-regular.woff');
```
janibonnevier: Where in the admin is this used?

BernhardPosselt: dead code, removed it

Issue #341, opened 2016-10-13
If an app is registered an /c/apps/APP_ID category should be created on help.nextcloud.com


BernhardPosselt:
To do this there need to be 3 additional settings in base.py

```
DISCOURSE_URL = 'https://help.nextcloud.com'
DISCOURSE_USER = None
DISCOURSE_TOKEN = None
```

If DISCOURSE_TOKEN is None, no requests should be made to the API. If a token is given, a new subcategory should be created in https://help.nextcloud.com/c/apps based on the app id e.g. https://help.nextcloud.com/c/apps/mail once a user registers an app.

https://meta.discourse.org/t/discourse-api-documentation/22706

PR #342, opened 2016-10-13
Remove discussion tag for good

### LINKED ISSUE #340: Ditch discussion tag
@georgehrke mentioned that it would be much more useful to have a full category on discourse so that you can mark questions as solved, see https://github.com/nextcloud/calendar/pull/123

We should ditch the discussion tag and instead use the discourse API to create a new category upon creating an app and link to the category instead. This would also improve info.xml compatibility with ownCloud.

@janibonnevier @nickvergessen @LukasReschke @Henni @ChristophWurst @MorrisJobke @nickvergessen

BernhardPosselt:
Fix #340 

Automatically creating categories is moved to #341


nextcloudappstore/core/api/v1/release/info.xsd
```diff
@@ -32,9 +32,6 @@
                             maxOccurs="1"/>
                 <xs:element name="repository" type="repository" minOccurs="0"
                             maxOccurs="1"/>
-                <xs:element name="discussion" type="discussion-url"
```
BernhardPosselt: If you validate it using the exact schema, it will. As for the app store: we've got some xslt in place that simply strips out unknown tags before validation.

PR #347, opened 2016-10-15
Allow changelog brackets

BernhardPosselt:
This also parses versions in square brackets correctly like described on http://keepachangelog.com/en/0.3.0/

@ChristophWurst


ChristophWurst:
👍 makes sens

PR #354, opened 2016-10-24
Only update the attributes on the app model if the uploaded app is ne…

BernhardPosselt:
…ither a pre-release nor a nightly, unless its the only existing release.

This keeps stuff like the description and other things "stable"

@janibonnevier @LukasReschke


nextcloudappstore/core/api/v1/release/importer.py
```diff
@@ -250,28 +252,46 @@ def _get_object(self, key: str, value: Any, obj: Any) -> Any:
...
-        for release in releases:
-            if uploaded_version < Version(release.version):
-                return False
-        return True
+        uploaded_version = Version(current_version)
+        is_prerelease = '-' in current_version
+        is_nightly = value['release']['is_nightly']
+        is_stable = not is_prerelease and not is_nightly
+
+        def newer_version(release: Any) -> bool:
```
janibonnevier: `is_newer_version` ?

PR #356, opened 2016-10-25
  By default exclude nightlies and pre-releases from release feed so 

BernhardPosselt:
users wont get spammed. If present add changelog to release feed

@janibonnevier

### DIFF
--- a/nextcloudappstore/core/feeds.py
+++ b/nextcloudappstore/core/feeds.py
@@ -3,6 +3,8 @@ from django.contrib.syndication.views import Feed
 from django.urls import reverse_lazy
 from django.utils.feedgenerator import Atom1Feed
 from django.utils.translation import ugettext_lazy as _  # type: ignore
+from parler.models import TranslationDoesNotExist
+
 from nextcloudappstore.core.models import AppRelease
 from markdown import markdown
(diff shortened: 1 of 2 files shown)


nextcloudappstore/core/feeds.py
```diff
@@ -13,14 +15,28 @@ class AppReleaseRssFeed(Feed):
...
     link = reverse_lazy('home')
 
+    def __call__(self, request, *args, **kwargs):
+        self.request = request
+        return super().__call__(request, *args, **kwargs)
+
     def items(self):
-        return AppRelease.objects.order_by('-last_modified')[:10]
+        queryset = AppRelease.objects.order_by('-last_modified')
+        if 'nightly' not in self.request.GET:
```
janibonnevier: Doesn't this mean that setting `?nightly=false` means "yes, include nightlies"? Same thing with `prerelease`.

BernhardPosselt: Yeah, but i dont think there's any particular big difference. You could also use ?nightly&prelrease

janibonnevier: OK, it's fine then :)

PR #357, opened 2016-10-30
Fix #131: Front end for app ownership transfer

### LINKED ISSUE #131: Allow transfer of app ownership

janibonnevier:
#131

This could probably use some polishing, but it seems to work and I think it should be reviewed before more work is put into it.

@BernhardPosselt


nextcloudappstore/core/forms.py
```diff
@@ -1,8 +1,10 @@
+from django.contrib.auth.models import User
```
janibonnevier: Would be nice to get the user model from the settings, but I'm not sure how to do that since `settings.AUTH_USER_MODEL` is a string.

BernhardPosselt: django.contrib.auth.get_user_model()

nextcloudappstore/core/forms.py
```diff
@@ -74,3 +76,31 @@ def save(self):
...
         app_rating.comment = self.cleaned_data['comment']
         app_rating.save()
+
+
+class AppOwnershipTransferForm(Form):
+    new_owner_username = CharField(
+        max_length=User._meta.get_field('username').max_length,
+        label=_('New owner (username)'))
+
+    def __init__(self, *args, app=None, **kwargs):
```
BernhardPosselt: I'd prefer: 
```py
def __init__(self, *args, **kwargs):
    self.app = kwargs.pop('app', None)
    super().__init__(*args, **kwargs)
```

Since I'm unsure if the variable arguments work as expected

nextcloudappstore/core/static/assets/js/ui/app-edit.js
```diff
@@ -0,0 +1,19 @@
...
+    'use strict';
+
+    // App ownership transfer form
+    let ownerTransferForm = document.getElementById('app-ownership-transfer-form');
+    let appId = ownerTransferForm.getAttribute('data-app-id');
+    let promptMsg = ownerTransferForm.getAttribute('data-prompt');
+    let failMsg = ownerTransferForm.getAttribute('data-fail');
+    ownerTransferForm.addEventListener('submit', (event) => {
+        event.preventDefault();
+        let appIdInput = window.prompt(promptMsg);
```
BernhardPosselt: Should be implemented similarly to the delete app form in the user profile. It's almost never a good idea to use alert or prompt

nextcloudappstore/core/templates/app/edit.html
```diff
@@ -0,0 +1,31 @@
...
+
+{% block head-title %}{% blocktrans with appname=object.name %}Edit app: {{ app_name }}{% endblocktrans %} - {% trans 'Apps' %} - {% endblock %}
+{% block head %}
+    <script defer="defer" src="{% static 'assets/js/ui/app-edit.js' %}"></script>
+    <meta name="language-code" content="{{ request.LANGUAGE_CODE }}">
+{% endblock %}
+
+{% block apps %}
+    <div class="header-wrap">
+        <h1>{% blocktrans with app_name=object.name %}Edit app: {{ app_name }}{% endblocktrans %}</h1>
```
BernhardPosselt: Should be **Edit app**. In general I'd avoid blocktrans since its harder to translate for translators.

nextcloudappstore/core/templates/app/edit.html
```diff
@@ -0,0 +1,31 @@
...
+    </div>
+    <p><a href="{% url 'app-detail' object.id %}">&larr; App details</a></p>
+    <div class="row">
+        <div class="col-md-12">
+            <section>
+                <h2>Transfer app ownership</h2>
```
BernhardPosselt: l10n

nextcloudappstore/core/templates/user/app-ownership-transfer.html
```diff
@@ -0,0 +1,42 @@
...
+
+{% block head-title %}{% trans 'App ownership transfers' %} - {% endblock %}
+
+{% block account-content %}
+    <h1>{% trans "Pending app ownership transfers" %}</h1>
+    <section>
+        {% for transfer in object_list %}
+            <h2><a href="{% url 'app-detail' transfer.app.id %}">{{ transfer.app.name }}</a></h2>
+            {% comment %}Translators: 'datetime' is rendered with the 'naturaltime' filter: https://docs.djangoproject.com/en/1.10/ref/contrib/humanize/{% endcomment %}
+            <p>{% blocktrans with datetime=transfer.proposed|naturaltime %}Transfer proposed {{ datetime }}{% endblocktrans %}</p>
```
BernhardPosselt: You can avoid the blocktrans and comment if you go for something like: 

```html
<p>{% trans 'Transfer proposed' %}: ...
```

nextcloudappstore/core/templates/user/app-ownership-transfer.html
```diff
@@ -0,0 +1,42 @@
...
+                {% if transfer.to_user == request.user %}
+                    <p><strong>{% blocktrans with other=transfer.from_user %}From {{ other }} to you{% endblocktrans %}</strong></p>
```
BernhardPosselt: Same here, simply go for Proposed from: user

nextcloudappstore/core/user/views.py
```diff
@@ -82,3 +86,40 @@ def get_context_data(self, **kwargs):
...
+
+
+class AppOwnershipTransferView(LoginRequiredMixin, ListView):
+    model = AppOwnershipTransfer
+    template_name = 'user/app-ownership-transfer.html'
+    http_method_names = ['get', 'post']
+
+    def post(self, request, *args, **kwargs):
+        transfer_id = request.POST.get('transfer-id', '')
+        op = request.POST.get('op', '')
```
BernhardPosselt: variable should be spelled out, was thinking for a second what the parameter should do

nextcloudappstore/core/user/views.py
```diff
@@ -82,3 +86,40 @@ def get_context_data(self, **kwargs):
...
+        user = request.user
+
+        if transfer_id and op:
+            transfer = self.model.objects.get(id=transfer_id)
```
BernhardPosselt: get_object_or_404?

nextcloudappstore/core/views.py
```diff
@@ -270,3 +273,42 @@ def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['form'] = AppRegisterForm()
         return context
+
+
+class AppEditView(LoginRequiredMixin, TemplateView):
```
BernhardPosselt: There's probably a better base class for an edit and detail view. Your current code needs to fetch the model from the context which is kinda meh and you also need to handle http methods and non existing objects with custom 404 code

nextcloudappstore/core/templates/user/app-ownership-transfer.html
janibonnevier: In that case I'd drop the naturaltime. As I see it, the point of it is to express things in natural language (without ":"), and therefore I chose blocktrans so that translators could have greater freedom in forming natural sentences in each of their languages.

nextcloudappstore/core/user/views.py
janibonnevier: Parameter too, or just the variable?

nextcloudappstore/core/views.py
janibonnevier: Maybe so. I think I started out with one, but switched to a more general view because I was unhappy about something, I'll try to figure out what.

nextcloudappstore/core/forms.py
```diff
@@ -74,3 +76,31 @@ def save(self):
         app_rating.set_current_language(self._language_code)
         app_rating.comment = self.cleaned_data['comment']
         app_rating.save()
+
+
+class AppOwnershipTransferForm(Form):
+    new_owner_username = CharField(
+        max_length=User._meta.get_field('username').max_length,
```
BernhardPosselt: A leading _ means private usually, why not reference the user model?

nextcloudappstore/core/user/views.py
BernhardPosselt: Same for parameter

nextcloudappstore/core/views.py
janibonnevier: Right, now I remember. I intended `AppEditView` to be able to support multiple forms (since the view is reached by clicking "Edit app"). Not sure if that is a good idea, but I suspect that views such as `UpdateView` only support one form in a meaningful way. Regardless, I will refactor this so that `post()` does not get stuff from the context.

janibonnevier:
Proper confirmation prompt is still missing, but I *think* I have fixed everything else.

nextcloudappstore/core/forms.py
```diff
@@ -89,3 +94,36 @@ def save(self):
         app_rating.set_current_language(self.cleaned_data['language_code'])
         app_rating.comment = self.cleaned_data['comment']
         app_rating.save()
+
+
+class AppOwnershipTransferForm(ModelForm):
+    to_user = UsernameField(label=_('New owner (username)'))
+    app_id_confirm = CharField(
+        label=_('App ID (for confirmation)'),
+        max_length=App._meta.get_field('id').max_length)
```
janibonnevier: `_meta` is not private (anymore), but a supported API: https://docs.djangoproject.com/en/1.10/ref/models/meta/

nextcloudappstore/core/views.py
```diff
@@ -297,3 +300,60 @@ def get_context_data(self, **kwargs):
...
```
janibonnevier: This view should not be shown to anyone else but the current owner of the app. This must be fixed if we choose to proceed with this approach.

BernhardPosselt:
Hi Jani, lets close this pull request and make it easier and more secure. I'm thinking of the following:

* In his profile settings an app maintainer sees a list of his apps. Each app has a button ("enable app transfer") that sets a flag on the App model to true (e.g. app.transfer_enabled = true)
* He can click on the button again ("disable app transfer") to reset the flag to false

The new owner just registers the app again by clicking on App developer -> register app, inserts the public certificate and the signature

The code in https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/views.py#L81 can then be changed to read:

```py
app = App.objects.get(id=app_id)
+  if not app.transfer_enabled and app.owner != request.user:
    msg = 'Only the app owner is allowed to update the certificate'
    raise PermissionDenied(msg)
app.certificate = certificate
+  app.owner = requets.user
app.save()
return Response(status=204)
```

That way you ensure that the new owner does in fact own the private key.

PS: if you don't feel like working on it, no worries I can take over the ticket :)

cc @LukasReschke

janibonnevier:
Alright :) You can take over, as I don't have much time for this.

BernhardPosselt:
@janibonnevier exams?

janibonnevier:
@BernhardPosselt No, those were last week. But next set of courses as well as work.

PR #364, opened 2016-11-04
Also store and expose number of ratings

### LINKED ISSUE #363: In addition to the rating the api should also expose the number of ratings
This should work similar to the current ratings, as in be cached on the app object. Default should be 0 and the value should only be updated once the threshold is reached.

BernhardPosselt:
Fix #363


BernhardPosselt:
@janibonnevier @LukasReschke @jancborchardt

PR #372, opened 2016-11-09
allow digits in app ids

### LINKED ISSUE #370: schema does not allow digits in app ids
Apparently app ids like ``twofactor_u2f`` are not allowed by the info.xml schema. Could we change that regex to allow digits too or are there any strong arguments against that?

ref https://github.com/nextcloud/appstore/blob/e4567ce707b332ca14eb35e322bff5ec4397191b/nextcloudappstore/core/api/v1/release/info.xsd#L245-L250

ChristophWurst:
should fix #370 but I didn't test it 🙈 🙊


nextcloudappstore/core/api/v1/release/info.xsd
```diff
@@ -245,7 +245,7 @@
     <!-- first level elements -->
     <xs:simpleType name="id">
         <xs:restriction base="limited-string">
-            <xs:pattern value="[a-z]+[a-z_]*[a-z]+"/>
+            <xs:pattern value="[a-z]+[a-z0-9_]*[a-z0-9]+"/>
```
go2sh: I would suggest something like `[a-z]_?([a-z0-9]+_?)*[a-z0-9]+`
This would prohibit name like  `a________________________1`

BernhardPosselt: Good point, also thought about this but should not be an issue. Sure you can do something like that but that would also be a valid app id. It's probably better to not restrict it harder than needed especially because app authors need to read the regex if it fails.

BernhardPosselt:
👍 from my POV if no objections from other people are left

BernhardPosselt:
Btw, @ChristophWurst could you add a test similar to this: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/tests/test_parser.py#L69 with an info.xml that contains your id here https://github.com/nextcloud/appstore/tree/master/nextcloudappstore/core/api/v1/tests/data/infoxmls ?

PR #376, opened 2016-11-16
Remove Legacy Dependency-Tag information from Docs

MariusBluem:
According to https://github.com/nextcloud/server/pull/2139 😉 ...we'll not longer need to use 9.2 💪 
@BernhardPosselt ; discussed with @LukasReschke @nickvergessen


MorrisJobke:
👍

PR #395, opened 2016-12-07
Filter incompatible platform versions in the API

BernhardPosselt:
@LukasReschke @janibonnevier @adsworth


BernhardPosselt:
Background:

If you query the current API, e.g. 

    curl http://localhost:8000/api/v1/platform/10.0.0/apps.json

All apps are returned that have at least one compatible release with the given platform, in this case 10.0.0.

Those apps are then serialized including all of their releases. This means that if you run the above query you would also receive releases that were only compatible with 11.0.0.

This pull request also filters all releases for the given platform version like it was intended.

PR #411, opened 2016-12-28
Properly return HTTP 400 instead of 500

### LINKED ISSUE #410: Error 500 upon publishing app via api
Please provide the following details for error reports. Ignore in case of a feature request.

## Details
When i publish my app (passman) , it fails with a error 500.
This started after i changed the description for my app in the app store.

Code i use for posting a new release:

```
    postData = {
        'download': 'https://releases.passman.cc/passman_' + _version + '.tar.gz',
        'signature': signature,
        'nightly': nightly
    }
    headers = {'Authorization': 'Token eddd2187bdaec7c05b67962f068500d1dc6c5e56',
               'Content-Type': 'application/json; charset=utf8'}

    r = requests.post('https://apps.nextcloud.com/api/v1/apps/releases', data=json.dumps(postData), headers=headers)
```

This worked before, as i said, the error 500 started showing up after i changed the [info.xml](https://github.com/nextcloud/passman/blob/master/appinfo/info.xml)

BernhardPosselt:
Fix #410 

Turns out APIException [returns a 500](https://github.com/tomchristie/django-rest-framework/blob/master/rest_framework/exceptions.py#L82) and not a 400 as intended.

@brantje @LukasReschke @adsworth @janibonnevier @MorrisJobke


brantje:
Thanks! Curious to see what is wrong with my app =)

BernhardPosselt:
["info.xml did not validate: Element 'summary': [facet 'maxLength'] The value has a length of '490'; this exceeds the allowed maximum length of '128'., line 13"]%

brantje:
Isn't 128 characters kinda low for an app description?

BernhardPosselt:
The summary is the small one liner that is rendered below the app thumbnail on the app list page. By default the description is the fallback, so if you don't add a summary and your description goes over 128 chars, it will be too long for the summary, check http://nextcloudappstore.readthedocs.io/en/latest/developer.html#info-xml

BernhardPosselt:
TL;DR: add a \<summary> element

brantje:
Thanks! Adding it now, will report back.

BernhardPosselt:
BTW, this is the xslt that is used to "migrate stuff" https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/release/pre-info.xslt

You also see that your other category is mapped to tools https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/api/v1/release/pre-info.xslt#L63 ;)

brantje:
Thanks, updated the summary, and category. 
I assume there is a caching system in front of the appstore?

BernhardPosselt:
Nope, we only use ETags for caching the JSON results.

brantje:
Strange, I've updated my app but changes are not visible in the appstore.

BernhardPosselt:
Pre-releases like RC1 don't update the page unless they are the only app release.

brantje:
So i'd have to bump version?

BernhardPosselt:
To a stable version right.

Issue #408, opened 2016-12-21
Default language fallback for comments if none present (Comments vanish?)


jancborchardt:
I left a comment at the Mail app with my rating: https://apps.nextcloud.com/apps/mail and there were also some other comments before.

Now there are none. Is this maybe due to a new version being released? If so, the comments should only indicate that they were made before the recent version. But definitely not deleted.

@nextcloud/appstore

BernhardPosselt:
Change your language, comments are only displayed in the language you've posted

jancborchardt:
Maybe if there are no comments in your language then English should be shown by default? Just to populate it a bit in the beginning since most people will comment in english.

BernhardPosselt:
That would probably be very confusing. If you post a comment in your language then one comment would be present and all English comments are gone. That's why you've filled your issue btw

jancborchardt:
No, I opened the issue because I posted the comment in english and I had my browser in english. Then I viewed it with a different browser set to french.

So basically everyone who has their browser set to a different language than English will think the appstore is basically devoid of other people in the beginning. That’s not a good first-run-experience.

As a different solution, how about for the first few weeks/months we just show _all_ comments regardless of the language? Most will be english anyway.

BernhardPosselt:
Providing a drop down to choose the language would solve this issue then I guess. The current language is preselected in the drop down and you can switch to a different language. 

BTW: Not many people use browsers in two different languages and logging in/choosing the language gets rid of that issue ;D

jancborchardt:
> Providing a drop down to choose the language would solve this issue then I guess.

No it wouldn’t.

> BTW: Not many people use browsers in two different languages and logging in/choosing the language gets rid of that issue ;D

That’s not my point. The point is that _any_ person using any other language than English will see no comments at all for any app. Because most if not all comments will be made in English.
So it seems like a dead appstore, which is not a good experience. So let’s show them all, at least for the first few weeks/months.

adsworth:
I think I'd currently agree with @jancborchardt that it's best to show other language comments rather than no comments.

How about this: If there are no comments in the users language we show a message "There are no comments in your language. Displayed are the comments in other languages" (or something to that effect) and then display the comments from all other languages.

This is a little heavier on the DB since we need at least two queries, but from a UX I think it's the best solution.

Spartachetto:
I'll put here my user experience just as a contribute to the decision.

[...]

PR #414, opened 2017-01-04
Comment language implements #408

adsworth:


nextcloudappstore/core/forms.py
```diff
@@ -50,27 +51,40 @@ class AppRegisterForm(Form):
                               'openssl base64</b>'))
     safe_help_fields = ['certificate', 'signature']
 
+def get_languages_local(language=None):
+    if language:
+        languages = list(language)
+    else:
+        languages = [ l[0] for l in settings.LANGUAGES]
+
+    language_infos = list()
```
BernhardPosselt: s/list()/[]  ?

nextcloudappstore/core/forms.py
```diff
@@ -50,27 +51,40 @@ class AppRegisterForm(Form):
...
+    for l in languages:
```
BernhardPosselt: looks like its super easy to replace with a list comprehension

nextcloudappstore/core/forms.py
```diff
@@ -50,27 +51,40 @@ class AppRegisterForm(Form):
                               'openssl base64</b>'))
     safe_help_fields = ['certificate', 'signature']
 
+def get_languages_local(language=None):
+    if language:
+        languages = list(language)
```
BernhardPosselt: [language] ?

nextcloudappstore/core/forms.py
```diff
@@ -50,27 +51,40 @@ class AppRegisterForm(Form):
...
+    else:
+        languages = [ l[0] for l in settings.LANGUAGES]
```
BernhardPosselt: This seems to be used in a choice field. Isn't a choice field a tuple of code and language?

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -13,7 +13,7 @@
     <script defer="defer" src="{% static 'assets/js/ui/app-detail.js' %}"></script>
     <meta name="nextcloudappstore-app-description-url" content="{% url 'app-description' object.id %}">
     <meta name="nextcloudappstore-app-ratings-url" content="{% url 'app-ratings' object.id %}">
-    <meta name="language-code" content="{{ request.LANGUAGE_CODE }}">
+    <meta name="language-code" content="{{ language_code }}">
```
BernhardPosselt: Should probably stay the same (since its the global language code). If this is actually the global language code we should bind it to the request using a middleware.

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -159,6 +159,24 @@ <h4 class="section-heading" id="comments">{% trans 'Discussion' %}</h4>
             </div>
         </div>
     </template>
+    <template id="app-rating-template-no-comments">
+        <div class="row">
+            <div class="col-md-12">
+                {% blocktrans %}No comments found.{% endblocktrans %}
```
BernhardPosselt: trans?

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -159,6 +159,24 @@ <h4 class="section-heading" id="comments">{% trans 'Discussion' %}</h4>
...
+            </div>
+        </div>
+    </template>
+    <div class="app-rating-language">
+        {% get_language_info_list for languages as languages %}
+        {% blocktrans %}Please choose the language to of the comments to display:{% endblocktrans %}
```
BernhardPosselt: trans?

nextcloudappstore/core/views.py
```diff
@@ -66,7 +65,19 @@ def post(self, request, id):
     def get_context_data(self, **kwargs):
         context = super().get_context_data(**kwargs)
         context['DISCOURSE_URL'] = settings.DISCOURSE_URL.rstrip('/')
-        context['rating_form'] = AppRatingForm()
+        language_code = get_language()
+        if not language_code:
+            language_code = self.request.LANGUAGE_CODE
+        context['rating_form'] = AppRatingForm(initial={'language_code':language_code})
+
+        rating_languages = set()
```
BernhardPosselt: = {} ?

nextcloudappstore/core/views.py
```diff
@@ -66,7 +65,19 @@ def post(self, request, id):
...
+        ratings = AppRating.objects.filter(app=context['app'])
+        for r in ratings:
```
BernhardPosselt: flatmap?

BernhardPosselt: flatmap() see https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/facades.py#L45

rating_languages = flatmap(lambda r: r.get_available_languages(), ratings)

nextcloudappstore/core/forms.py
adsworth: Yes, see https://github.com/nextcloud/appstore/pull/414#discussion_r94726283 where I'm getting the local name to display in the choice field.

nextcloudappstore/core/forms.py
adsworth: but I can't come up with a list comprehension without calling get_language_info(l) twice for each language, which I don't think is a good idea either.

BernhardPosselt: Ah right. You'd need a tmp variable for lang_info

BernhardPosselt: [(i['code'], i['name_local']) for i in [get_language_info(l) for l in languages]]

nextcloudappstore/core/forms.py
```diff
@@ -50,27 +51,40 @@ class AppRegisterForm(Form):
...
     safe_help_fields = ['certificate', 'signature']
 
+def get_languages_local(language=None):
+    if language:
+        languages = [language]
+    else:
+        languages = [ l[0] for l in settings.LANGUAGES]
+
+    language_infos = []
+    for l in languages:
```
BernhardPosselt: [(i['code'], i['name_local']) for i in [get_language_info(l) for l in languages]]

adsworth:
@BernhardPosselt it implements 

- posting in a different language 
- display a message if no comments were found for request.LANGUAGE_CODE
- allow the user to choose the language of the comments displayed.

BernhardPosselt:
I think you've accidentally commited an sqlite db. 

Apart from that the code looks good, checked out the ui and came across the following things:

* The comment language select label should read "Display comments in:"
* If no rating comments were made, the dropdown is empty. Instead it should just be hidden including the label
* The "No comments found." should be more promiment so that you don't confuse it with the content itself (bigger font, centered?) and it should have some top and bottom padding (20-30px?)
* Posting a comment in English and then switching to German in your profile will display no comments while the selected language is in the drop down is still English. Instead of available languages we should just render all languages into that drop down (settings.LANGUAGES)
* Language code label should read: Language
* Language select should be before the textarea so its clear that it applies to the textarea. Also you read it first and it might prevent your comment being deleted if change the language afterwards (because you didnt notice it before)

Points that are probably good to work on in another pull request/ticket:
* On initial page load the comment language does not default to English if no comments are present in the user's language
* Changing the post language should retrieve the actually previously posted content and fill it into the textarea (Unless there is none or its empty). Think of it like this: I comment in English, German and French. Then I want to modify my German comment. I should be able to modify it by changing the language drop down to German and then edit my previously created comment. We probably need to disable the text area while loading the previous content

adsworth:
@BernhardPosselt  "...Posting a comment in English and then switching to German in your profile will display no comments while the selected language is in the drop down is still English. Instead of available languages we should just render all languages into that drop down (settings.LANGUAGES)..."

I disagree on that point what's the point of giving the user a choice that is totally useless for him since there ar eno comments?

adsworth:
I added section-headings for the Rate form and for the Comments as a visual distinction.

BernhardPosselt:
> I disagree on that point what's the point of giving the user a choice that is totally useless for him since there are no comments?

My guess is that you want to show the user in which languages there are comments so he doesn't bother looking into the other languages. You can achieve this in multiple ways. One would be to display the count left to the language if there is one, e.g. English (12), German, French (2). Or you could add an empty option that is selected if there are no comments for the user's language

My main goal here is: you wanted to view language X but there are no comments. Currently Language X will be the first in the drop down which is confusing.

BernhardPosselt:
PS: you can tick off the check boxes in my previous post if you think it can be reviewed :D

adsworth:
@BernhardPosselt how does the latest change work for you? I'm now making sure the sessions language is in the language select, even if there are no comments.

BernhardPosselt:
@adsworth yep, looks good

PR #418, opened 2017-01-06
Slider fixes

BernhardPosselt:
Regression fixes from #415 

The JavaScript code expects the slider buttons to be present, otherwise it fails with errors. This manifests itself when you try to click the full screen button.

You can test it out by clicking on the full screen button or observing the js console here https://apps.nextcloud.com/apps/ojsxc

Apart from that a few cosmetic changes:

* pyvenv command is deprecated in Python 3.6
* make admin nav easier to read and distinguishable from the default page by not using a bg image

@adsworth

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
   - "3.4"
   - "3.5"
   - "3.6"
-  
+
 env:
   matrix:
     - DB=postgres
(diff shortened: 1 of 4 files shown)


adsworth:
how about hiding them via js?
Something like (warning untested code from a Javascript noob) :)

```
if $('.img-strip img').length <= 1 ) { $('.img-slider-controls').hide(); }
```

BernhardPosselt:
The JS code needs a bit of a clean up at some point so I don't really want to touch it for now :)

BernhardPosselt:
Stuff like it shouldnt even fail if there are no buttons.

BernhardPosselt:
Opened https://github.com/nextcloud/appstore/issues/419 fyi

adsworth:
looks good. +1

PR #420, opened 2017-01-07
Fix django debug toolbar and expose HTML gui for REST api

BernhardPosselt:
There were some changes in the toolbar code that require additional setup


BernhardPosselt:
@LukasReschke @adsworth @janibonnevier

PR #423, opened 2017-01-08
Rating fixes

BernhardPosselt:
and very small adjustments that I forgot in the review :)

* We don't use jQuery since it's outdated. For selecting elements there's querySelector and querySelector all, for ajax, get, post etc there's fetch. More info here http://youmightnotneedjquery.com/
* I changed comment to rating comment to fit the overall language. Using comment encourages users to post enhancement and bug reports.
* I've added a label around the comment selector text to fit the overall style

cc @adsworth


adsworth:
looks good. +1

PR #424, opened 2017-01-14
Fix layout issues with long names

BernhardPosselt:
Add an ellipsis for names too and allow to read the full text by using title 

@nickvergessen @adsworth @LukasReschke @janibonnevier


nickvergessen:
👍

Issue #422, opened 2017-01-08
 Continuation from #414: view and edit your own comments in different languages


BernhardPosselt:
Changing the post language should retrieve the actually previously posted content and fill it into the textarea (Unless there is none or its empty). Think of it like this: I comment in English, German and French. Then I want to modify my German comment. I should be able to modify it by changing the language drop down to German and then edit my previously created comment. We probably need to disable the text area while loading the previous content

Furthermore if the text area is empty the button should not show "Update rating"

BernhardPosselt:
Fixed

PR #433, opened 2017-02-06
WIP:allow commenting in different languages. Implements #422

adsworth:
when switching language in the rating form loads the current users comment in the newly selected language. What is open is showing a dialog, if there is unsaved data, when switching language asking the user if he want's to save.


nextcloudappstore/core/static/assets/js/ui/app-detail.js
```diff
@@ -41,6 +41,16 @@
...
     }
 
+    function load_language(lang) {
+        commentTextarea.readOnly = true;
+        fetch(commentUrl + "?lang=" + lang, {credentials: 'include'})
+            .then((response) => response.json())
+            .then((json) => {
+                commentTextarea.value = json.comment;
+        });
+        commentTextarea.readOnly = false;
```
BernhardPosselt: This part will execute after you fired the request, not when it finished successfully, see https://github.com/github/fetch/issues/223#issuecomment-148927226

nextcloudappstore/core/views.py
```diff
@@ -45,6 +52,24 @@ def get_queryset(self):
             app=app)
 
 
+class AppCommentApi(View):
+
+    def get(self, request, *args, **kwargs):
```
BernhardPosselt: I'm unsure if you can just extend a DetailView and reuse more code, just take a look ;)

BernhardPosselt:
One thing I was wondering about btw: if a user changes the language drop down and he changed the content, should we ask him to confirm to drop changes?

adsworth:
I guess we have to ask for some kind of cinformation, otherwise we are going to loose data.
Can you point me somewhere so I can implement a nice modal popup asking for confirmation.
And then the question is do we submit the new comment via an API or do we just let the user submit the form?

BernhardPosselt:
As for modals: simply use the built in (confirm()), shouldn't be to bad in this case :)
Comment submission should be via plain form

BernhardPosselt:
Moved most of your code into https://github.com/nextcloud/appstore/pull/455/files 

Let's not brood too much about the current situation, let's first get in a working solution and work on UX later once we've got a good enough idea of how we should actually implement it.

PR #480, opened 2017-05-02
Validate database.xml files

BernhardPosselt:
@nickvergessen @LukasReschke @MorrisJobke @georgehrke @adsworth @jospoortvliet @ChristophWurst


nickvergessen:
We should do this with the app checker...

ChristophWurst:
> We should do this with the app checker...



Having both would also make sense. Then you'd catch schema violations while development or on your CI system *before* you actually release the app.

nickvergessen:
I added it to my todo list

PR #489, opened 2017-06-10
Remove nextcloud 9 and 10 compatibility

BernhardPosselt:
- Dropped support for Nextcloud 9 and 10:
  - owncloud tag in info.xml will not be migrated to nextcloud tags anymore
  - ocsid will not be parsed anymore and was removed from the info.xsd
  - old categories will not be migrated anymore (tool, game, productivity, other)
  - presence of owncloud tag will not be validated anymore for apps depending on 9 and 10
  - v0 API was removed
  - nextcloudrelease.json fixtures for 9 and 10 were removed

@MorrisJobke @nickvergessen @LukasReschke


nickvergessen:
IIRC there will be one more release for 10, but luk may know more.

CHANGELOG.md
```diff
@@ -5,6 +5,13 @@
 ### Removed
 
 - Dropped Node.js 7 support and require Node.js 8+
+- Dropped support for Nextcloud 9 and 10:
+  - owncloud tag in info.xml will not be migrated to nextcloud tags anymore
+  - ocsid will not be parsed anymore and was removed from the info.xsd
+  - old categories will not be migrated anymore (tool, game, productivity, other)
+  - presence of owncloud tag will not be validated anymore for apps depending on 9 and 10
```
nickvergessen: this means when an app only has the oc tag, nextcloud 13 will suggest it to install? :/

BernhardPosselt: No, existing apps will continue to work but new uploads for 9 and 10 will not have all checks done that ensure that you can install them on 9 and 10

BernhardPosselt: And of course they will be required to add a nextcloud tag now because we don't guess it from the owncloud one any more

PR #502, opened 2017-07-29
Scrap auth category and go with security instead

### LINKED ISSUE #375: Rename Authentication & authorization
**Authentication & authorization** is a bit to long as a name for this category ... It does not really fit into the list if you have a look into the short other names like **Files** or **Multimedia** 😜 

@BernhardPosselt

BernhardPosselt:
fix #375


BernhardPosselt:
More details on this change: 
* items in auth category will be migrated to the security category
* both tags are still available in the info.xml but auth will be migrated to security

PR #521, opened 2017-09-18
add collaboration type

blizzz:
Depends on https://github.com/nextcloud/server/pull/6328, used in https://github.com/nextcloud/circles/pull/126

It looks fairly broad and complex… because "collaboration" is a pretty wide term and I've no idea what else could come there in future (or not). 

``shareType`` (line 461) could be an enumeration as well, however it would mean naming constants here and have a coupling to code. Not sure this is necessary, but if worthwhile I can change this. Opinions welcome about it.

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -277,6 +277,11 @@ A full blown example would look like this (needs to be utf-8 encoded):
                 <type>link</type>
             </navigation>
         </navigations>
+        <collaboration>
+            <plugins>
+                <plugin type="collaborator-search" shareType="SHARE_TYPE_CIRCLE">OCA\Circles\Collaboration\v1\CollaboratorSearchPlugin</plugin>
+            </plugins>
+        </collaboration>
     </info>
 
 The following tags are validated and used in the following way:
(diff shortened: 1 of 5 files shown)


BernhardPosselt:
Needs a description of what the tag actually does here in the docs: http://nextcloudappstore.readthedocs.io/en/latest/developer.html#info-xml

See https://github.com/nextcloud/appstore/blob/master/docs/developer.rst

As for the structure: the nesting seems pretty extreme given that you only want to be able to register search plugins. Maybe its possible to bring it down to something like (just some playing around, do not know the usecase enough for that):

```xml
<collaboration>
    <plugin section="search" type="SHARE_TYPE_CIRCLE">OCA\Circles\Collaboration\v1\CollaboratorSearchPlugin</plugin>
</collaboration>
```
As for the share type: is there a finite number of types? If so, use an enumeration.

blizzz:
@BernhardPosselt ty! Your remarks are addressed

BernhardPosselt:
Very small nitpick: XML and HTML usually uses **an-attribute** instead of **anAttribute**. Feel free to change it if you want to ;)

blizzz:
> Very small nitpick: XML and HTML usually uses an-attribute instead of anAttribute. Feel free to change it if you want to ;)

Uff… did not pay attention to this. Perhaps, also because _minOccurs_, _maxOccurs_ or _xsi:noNamespaceSchemaLocation_ for instance do not  follow this style as well. I acknowledge conventions and consistency however → will change this.

PR #526, opened 2017-10-13
Update models.py

rakekniven:
Wording changes due to discussions at Transifex.
https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/appstore/98030072

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -594,8 +594,8 @@ class AppReleaseDeleteLog(Model):
     last_modified = DateTimeField(auto_now=True, db_index=True)
 
     class Meta:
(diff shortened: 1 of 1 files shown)


BernhardPosselt:
Hm, doesn't really fit that well. This stuff is about keeping track of deleted app releases so we can update the http etag properly

rakekniven:
To be honest we do not fully understand the meaning of the original strings.

BernhardPosselt:
Do you think you can figure out the meaning with my description?

comradekingu:
@BernhardPosselt So, deleted nextcloud apps?

BernhardPosselt:
Deleted nextcloud app releases

PR #530, opened 2017-10-29
Added code signing to Makefile #529

### LINKED ISSUE #529: Add code signing to Makefile
It would be very nice to have appstore code signing added to the app-templates Makefile. Started it here: https://github.com/nextcloud/polls/blob/master/Makefile.

splitt3r:
I think this way the Travis CI file looks way less cryptic for beginners and signing the appstore package gets very easy. Feel free to update the code / beautify it. Fixes #529.


BernhardPosselt:
Code signing in that form is actually more harmful than just signing the package because you can't get rid of it (look at the recent outcry after a contacts release which failed to include one file and broke during upgrade).

So I'd not add that.

 The nextcloud ci scripts just add another level of failure and don't really add that much since the file is already pretty slim. 

Please open an issue next time before you invest any work ;)

BernhardPosselt:
Currently sick in bed so sry for short response let's discuss this again in a few days

splitt3r:
Ok, no problem. I already created in issue #529.

PR #547, opened 2018-02-07
Allow personal settings via info.xml and doc links

nickvergessen:
Ref https://github.com/nextcloud/server/pull/8232


nextcloudappstore/api/v1/release/info.xsd
```diff
@@ -357,6 +364,10 @@
                         maxOccurs="unbounded"/>
             <xs:element name="admin-section" type="php-class" minOccurs="0"
                         maxOccurs="unbounded"/>
+            <xs:element name="personal" type="php-class" minOccurs="0"
```
nickvergessen: Supported with 13+
https://github.com/nextcloud/server/pull/4890

nextcloudappstore/api/v1/release/info.xsd
```diff
@@ -242,6 +242,13 @@
         </xs:restriction>
     </xs:simpleType>
 
+    <xs:simpleType name="doc-url">
+        <xs:restriction base="xs:anyURI">
+            <xs:pattern value="(https://.+|(admin|user|developer)-[a-z]+[\-a-z]*[a-z]+)"/>
```
nickvergessen: Supported with 7+ 🙈

BernhardPosselt: base="xs:anyURI" is now wrong, should be base="non-empty-string"

nextcloudappstore/api/v1/release/info.xsd
BernhardPosselt: also updating the docs would be nice https://github.com/nextcloud/appstore/blob/master/docs/developer.rst#app-metadata

E.g. what interface to implement or which base class to extend and what it does

BernhardPosselt:
I will take care of the tests and parsing. I suppose we just ignore non urls on the store?

BernhardPosselt:
Your regex does not work btw, test fails ;)

BernhardPosselt:
Is there a use case for having 

    <admin>user-something</admin>

nextcloudappstore/api/v1/release/info.xsd
BernhardPosselt: **admin-** prefix should probably only work in \<admin> tags right?

nickvergessen: Well it's not enforced on server side, but yeah it's the case at the moment, so I guess we can do that.

nickvergessen:
Btw, when you changed the info.xsd, please leave a note at https://github.com/nextcloud/server/pull/8232 or update the file directly.

BernhardPosselt:
@nickvergessen adjusted the schema for you, now only needs docs

nextcloudappstore/api/v1/release/info.xsd
nickvergessen: done

PR #567, opened 2018-07-17
+ dashboard/widget

daita:
add
 ```
 <dashboard>
   <widget>PhpClass</widget>
 </dashboard>
```
However, not sure those changes are enough !

### CODE REVIEW

**Comment by BernhardPosselt** [comment]
Hi @daita 

Not sure if we should fill up an apps info.xml with meta information for other apps. This makes future development harder for the server (e.g. think about nextcloud adding a dashboard default app and wanting to use the same tag).

I suppose your use case is adding/fetching dashboard plugins correct? In that case, why don't you use something different than the info.xml file? There are so many ways like dropping a file in a certain path (e.g. lib/Dashboard/Plugin.php) or use the database to register plugins.

**Comment by daita** [comment]
Well, this is supposed to be the nextcloud dashboard default app.

I will check with the @nextcloud/designers for other solutions.

**Comment by daita** [comment]
@MorrisJobke - please halp :)

Branch: https://github.com/nextcloud/dashboard/tree/gridstack
PR on server: https://github.com/nextcloud/server/pull/10222
Widgets on their way: https://github.com/nextcloud/dashboard/issues/39

Could be great to have this for NC14.

**Comment by MorrisJobke** [comment]
> Could be great to have this for NC14.

It's a bit late for that ... freeze was on Friday ... at least for the server PR.

**Comment by MorrisJobke** [comment]
As written in the server ticket: let's aim for the Nextcloud 15 release.

**Comment by BernhardPosselt** [comment]
Do we want to add extra tags for non core apps as well? Maybe makes sense to add a non schema validated section like:

```xml
<meta>
    <! -- all elements here allowed here using https://www.w3schools.com/XML/schema_complex_any.asp -->
</meta>
```
So OP could add

```xml
<meta app="dashboard">
    <widget>PhpClass</widget>
</meta>
```

But I think using a file name by convention or storing the setting in the database is ultimately a better idea because OP can simply query the database or include a file if it exists without having to parse all app's xml again (given that the server does not persist this on install)

daita:
@BernhardPosselt @MorrisJobke 

What about this PR, now that the interface for the dashboard are in the core of Nextcloud ?

Issue #578, opened 2018-09-14
Prepare app store for delivering what's new information


blizzz:
In order to add support for "What's new" information for apps (https://github.com/nextcloud/server/issues/9165), we would like to extend the appstore with

* [ ] an XML schema to provide that information (essentially https://github.com/nextcloud/changelog_server/blob/master/schema.xsd)
* [ ] the ability to provide this information in the requested language alongside other meta data in the corresponding api call(s)

bdovaz:
Any news? I would like to use it with Renovate.

Thank you!

PR #580, opened 2018-09-24
🚧🏗👷 Accept and deliver What's New information

blizzz:
:construction: WIP solves #578 :construction:


BernhardPosselt:
What about simply parsing https://keepachangelog.com/en/1.0.0/

blizzz:
> What about simply parsing https://keepachangelog.com/en/1.0.0/ which is already supported by the store?

That's too detailed. The idea is to provide up to 3 easy-understandable, non-technical items for end users (plus three for admins). Sort of release highlights.

BernhardPosselt:
Have you thought about using dashes instead of camel case? e.g. \<whats-new> instead of \<whatsNew> etc?

nextcloudappstore/api/v1/release/changes.xsd
```diff
@@ -0,0 +1,55 @@
...
+                </xs:element>
+            </xs:sequence>
+            <xs:attribute name="version" type="version" use="required" />
+        </xs:complexType>
+    </xs:element>
+
+    <xs:simpleType name="version">
+        <xs:restriction base="xs:string">
+            <xs:pattern value="(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)"/>
+            <xs:minLength value="6"/>
```
BernhardPosselt: I'd drop the min and max length requirements. min length of 6 makes 0.0.1 invalid btw.

nextcloudappstore/api/v1/release/changes.xsd
```diff
@@ -0,0 +1,55 @@
...
+            <xs:maxLength value="12"/>
+        </xs:restriction>
+    </xs:simpleType>
+
+    <!-- based on xs:language, just uses _ as a separator between lang and region. -->
+    <xs:simpleType name="language">
+        <xs:restriction base="xs:token">
```
BernhardPosselt: Should make be consistent with https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/api/v1/release/info.xsd#L136

nextcloudappstore/api/v1/release/changes.xsd
```diff
@@ -0,0 +1,55 @@
...
+            <xs:sequence>
+                <xs:element name="changelog" maxOccurs="1" minOccurs="1">
+                    <xs:complexType>
+                        <xs:attribute name="href" type="xs:anyURI" use="required"  />
+                    </xs:complexType>
+                </xs:element>
+                <xs:element name="whatsNew" minOccurs="1" maxOccurs="unbounded">
+                    <xs:complexType>
+                        <xs:sequence>
+                            <xs:element name="regular" type="whatsNewInfo" minOccurs="1" maxOccurs="1" />
```
BernhardPosselt: What if there are only admin changes? like "ui for configuring additional mount type"

nextcloudappstore/api/v1/release/provider.py
```diff
@@ -35,6 +35,9 @@ def get_release_info(self, url: str, is_nightly: bool = False) -> Release:
             info = parse_app_metadata(meta.info_xml, self.config.info_schema,
                                       self.config.pre_info_xslt,
                                       self.config.info_xslt)
+            if meta.changes_xml:
+                whats_new = parse_app_whats_new(meta.changes_xml, self.config.changes_schema,
```
BernhardPosselt: Looks like an unused variable. If it's just for validation simply remove the assignment

nextcloudappstore/scaffolding/app-templates/12/app/appinfo/changes.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="utf-8"?>
+<release xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:noNamespaceSchemaLocation="https://apps.nextcloud.com/schema/apps/changes.xsd" version="0.0.1">
+    <changelog href="" />
+    <whatsNew lang="en">
+        <regular>
+            <item>First {{ app.name }} release!</item>
+        </regular>
+        <admin>
+            <item>requires {{ app.nextcloud_version }}</item>
```
BernhardPosselt: This is redundant with the version in info.xml and probably a bad example for new developers. Better example or simply remove it?

nextcloudappstore/api/v1/release/changes.xsd
```diff
@@ -0,0 +1,55 @@
+<?xml version="1.0"?>
+<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
+           elementFormDefault="qualified" attributeFormDefault="unqualified">
+    <xs:element name="release">
+        <xs:complexType>
+            <xs:sequence>
+                <xs:element name="changelog" maxOccurs="1" minOccurs="1">
```
BernhardPosselt: This could be optional and re-use the CHANGELOG.md in the app folder by default
.

nextcloudappstore/api/v1/release/changes.xsd
blizzz: Yeah, we should consider that. Might make sense for the server, too, as the maintenance releases often don't have much to brag about, but subtle bug fixes.

nextcloudappstore/api/v1/release/provider.py
blizzz: still WIP

blizzz:
fyi, i was sick and need to reschedule a bit. so, will be a bit until i pick it up again.

alexbijma:
hi, is there any progression in this? This would be very very useful to have.

PR #585, opened 2018-10-23
Allow arbitrary order of elements in info.xml

ChristophWurst:
@rullzer as discussed

### DIFF
--- a/nextcloudappstore/api/v1/release/info.xsd
+++ b/nextcloudappstore/api/v1/release/info.xsd
@@ -4,7 +4,7 @@
 
     <xs:element name="info">
         <xs:complexType>
-            <xs:sequence>
+            <xs:all>
                 <xs:element name="id" type="id" minOccurs="1" maxOccurs="1"/>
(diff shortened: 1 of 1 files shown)


juliushaertl:
That won't work as xs:all does only allow elements with maxOccurs of 0 or 1.

ChristophWurst:
Do you have any references for that?

https://stackoverflow.com/questions/16101488/difference-between-xsdall-and-xsdsequence-in-schema-definition#comment23005142_16101544

juliushaertl:
Ah I wasn't aware that it is supported in XSD 1.1, anyway seems the library lxml used by the appstore doesn't support that properly, see the failing tests.

BernhardPosselt:
Needs xsd 1.1 unfortunately which is only supported on Java and C# platforms as far as I know.

ChristophWurst:
Dang it.

BernhardPosselt:
Saxon C supports it http://www.saxonica.com/saxon-c/release-notes.xml

but I don't think it's integrated widely yet or available as packages on major distros.

PR #586, opened 2018-11-01
add the tag fulltextsearch

daita:


daita:
in relation of the implementation of the public interface of fulltextsearch within NC15:

https://github.com/nextcloud/server/pull/12195

BernhardPosselt:
You forgot to define where the element is expected

daita:
@BernhardPosselt My bad, doing this right now !

Issue #546, opened 2018-02-06
Notify users via email after password was changed


nickvergessen:
To notify users when they forgot to log out somewhere and someone else took over their account.

BernhardPosselt:
Can be solved by adding another line here https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/user/signals.py#L32 

Sending mails to a user is documented here: https://docs.djangoproject.com/en/2.0/topics/email/

mario:
This looks done so closing.

PR #592, opened 2019-01-14
added notify user on password change, issue #546

woopsix:


nextcloudappstore/user/signals.py
```diff
@@ -26,9 +27,20 @@ def password_changed_signal(sender, instance, **kwargs):
     :param kwargs:
     :return:
     """
+    from_mail = 'appstore@nextcloud.com'
```
BernhardPosselt: This should probably be moved into the settings/base.py

nextcloudappstore/user/signals.py
```diff
@@ -26,9 +27,20 @@ def password_changed_signal(sender, instance, **kwargs):
...
+    mail_subect = 'Nextcloud Appstore password changed'
```
BernhardPosselt: Both messages should be translated

woopsix: since i'm not familiar with the translation, could you please provide a place so i can read about it?

BernhardPosselt: import here: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/models.py#L16

usage here: https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/core/models.py#L60

nextcloudappstore/user/signals.py
```diff
@@ -4,6 +4,9 @@
 from django.db.models.signals import post_save, pre_save
 from django.dispatch import receiver
 from django.core.mail import send_mail
+from django.utils.translation import gettext as _
+
+from nextcloudappstore.settings.base import NEXTCLOUD_EMAIL
```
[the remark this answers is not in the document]
woopsix: Thanks a lot for the hints.

BernhardPosselt:
Thanks!

BernhardPosselt:
BTW, pushed some docs on it and changed *NEXTCLOUD_FROM_EMAIL* to the already included by default *DEFAULT_FROM_EMAIL*

PR #601, opened 2019-03-08
Provide an API endpoint to return all apps

### LINKED ISSUE #600: Need to return all apps on >=NC16
For https://github.com/nextcloud/server/pull/14578

We to deliver the full app list when on NC 16. No server side filtering as the admin gets the option to force enable the app.

CC: @MorrisJobke @blizzz

MorrisJobke:
* same as `GET /api/v1/platform/{**platform-version**}/apps.json` but without the platform version filter
* fixes #600

### DIFF
--- a/nextcloudappstore/api/v1/views.py
+++ b/nextcloudappstore/api/v1/views.py
@@ -28,6 +28,18 @@ from nextcloudappstore.core.throttling import PostThrottle
 from nextcloudappstore.core.versioning import version_in_spec
 from nextcloudappstore.user.facades import update_token
 
+APP_PREFETCH_LIST = [
+    'authors',
+    'screenshots',
+    'categories',
+    'translations',
+    'releases__translations',
+    'releases__phpextensiondependencies__php_extension',
+    'releases__databasedependencies__database',
+    'releases__shell_commands',
+    'releases__licenses',
+]
+
 
 class CategoryView(ListAPIView):
     queryset = Category.objects.prefetch_related('translations').all()
@@ -44,6 +56,11 @@ class NextcloudReleaseView(ListAPIView):
     serializer_class = NextcloudReleaseSerializer
 
 
+class AppsView(ListAPIView):
+    queryset = App.objects.prefetch_related(*APP_PREFETCH_LIST).all()
+    serializer_class = AppSerializer
+
+
 class AppView(DestroyAPIView):
     authentication_classes = (authentication.TokenAuthentication,
                               authentication.BasicAuthentication,)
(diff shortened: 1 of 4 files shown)


nextcloudappstore/api/v1/views.py
```diff
@@ -44,6 +44,11 @@ class NextcloudReleaseView(ListAPIView):
     serializer_class = NextcloudReleaseSerializer
 
 
+class AppsView(ListAPIView):
+    queryset = App.objects.all()
```
BernhardPosselt: You want to add prefetching, otherwise this will create thousands of SQL queries (n+1 problem), see https://github.com/nextcloud/appstore/blob/master/nextcloudappstore/api/v1/views.py#L56 for a list of required relations that should be prefetched. See https://docs.djangoproject.com/en/2.1/ref/models/querysets/#prefetch-related and possibly https://docs.djangoproject.com/en/2.1/ref/models/querysets/#select-related

nextcloudappstore/core/caching.py
```diff
@@ -33,6 +33,10 @@ def app_rating_etag(request: Any, id: str) -> str:
     return create_etag([(AppRating.objects.filter(app__id=id), 'rated_at')])
 
 
+def all_apps_etag(request: Any) -> str:
+    return create_etag([(App.objects.all(), 'last_modified')])
```
BernhardPosselt: Would also include the app release delete log like in apps_etag. In fact you should be able to reuse the function without changes

nextcloudappstore/api/v1/views.py
BernhardPosselt: You could extract that list into a global/singleton btw and just reuse it in both places.

MorrisJobke: Did it like that - could you check again?

nextcloudappstore/core/caching.py
MorrisJobke: Updated. Is this how you imagined it?

nextcloudappstore/core/caching.py
```diff
@@ -33,6 +33,13 @@ def app_rating_etag(request: Any, id: str) -> str:
     return create_etag([(AppRating.objects.filter(app__id=id), 'rated_at')])
 
 
+def all_apps_etag(request: Any) -> str:
```
BernhardPosselt: this is the same code as apps_etag right?

BernhardPosselt: just use apps_etag, it works for both routes

MorrisJobke: 🙈 I didn't noticed that - you are right - fixed

MorrisJobke:
@BernhardPosselt Fine with you?

MorrisJobke:
Okay - CI fails - let's look into this https://travis-ci.org/nextcloud/appstore/jobs/507923223

MorrisJobke:
I fixed the one issue with the list vs tuple and the code style - this is now ready for the final review IMO. 🚀

MorrisJobke:
The separate tag function is needed, because the method does not have a `version` parameter, that is required by `apps_etag`:

```
Traceback (most recent call last):
  File "lib/python3.5/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "lib/python3.5/site-packages/django/core/handlers/base.py", line 126, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "lib/python3.5/site-packages/django/core/handlers/base.py", line 124, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "lib/python3.5/site-packages/django/views/decorators/http.py", line 88, in inner
    res_etag = etag_func(request, *args, **kwargs) if etag_func else None
TypeError: apps_etag() missing 1 required positional argument: 'version'
```

PR #614, opened 2019-08-13
Fix blind SSRF for new releases

mario:


BernhardPosselt:
Looks good from a first glance, can you elaborate the issue and fix in code comments since it's security relevant?

@MorrisJobke you may want to stay updated on this as well

mario:
@BernhardPosselt absolutely. done.

BernhardPosselt:
Currently the build fails because of linting errors (see make lint). If the suite passes it's good to go

PR #617, opened 2019-08-14
Improve app  name validation UX

mario:
Previously we've had the same message shown twice upon failed validation. This is no longer true.

### DIFF
--- a/.idea/misc.xml
+++ b/.idea/misc.xml
@@ -3,7 +3,5 @@
   <component name="JavaScriptSettings">
     <option name="languageLevel" value="ES6" />
   </component>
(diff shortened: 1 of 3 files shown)


BernhardPosselt:
Good catch. I'm not sure if removing the hint about camel casing is a good idea. It should be obvious from the start what kind of pattern is expected. 

Apparently bootstrap also highlights the .help-block class in red if it's inside a .has-error element and therefore it looks as if the error message is present twice.

Maybe move the description to the label instead? Like: "App name (in CamelCase)"

mario:
Updated.

BernhardPosselt:
Can be merged after a successful build :+1:

PR #618, opened 2019-08-14
Prevent guessing users

mario:
It is currently possible to find valid users via the password reset form. This prevents it.

Do note that this could use more polishing on the UX side, but let's merge this as a first step.


BernhardPosselt:
Doesn't the reset password form simply tell you which email addresses are registered? You need a separate user name for the login.

mario:
@BernhardPosselt sure, but why would you want to expose that?

BernhardPosselt:
Right, I think it's best if nothing is exposed at all but I guess trying a bunch of existing E-Mails doesn't really give you a lot to work with.

We should probably fix this upstream at django-allauth since it's a low impact thing IMHO rather than handrolling a custom implementation on our own which potentially makes it harder to upgrade.

mario:
Generally I'm fine with that, but the question remains if they'd accept such a change or not.

BernhardPosselt:
Sure, I'd just ask first on their tracker if they are interested.

BernhardPosselt:
https://github.com/pennersr/django-allauth/issues/2341

mario:
Thx.

blizzz:
Since upstream is not really active, I guess we merge this here for so long?

BernhardPosselt:
@blizzz yeah, maybe add some comments in the source code like "remove this class once issue #xxx is resolved django-allauth" so it won't be forgotten.

mario:
@BernhardPosselt do it! :)

blizzz:
Not sure it is related, but the tests are failing now. @mario can you have a look, please?

BernhardPosselt:
@blizzz problem is https://github.com/nextcloud/appstore/issues/622

BernhardPosselt:
someone needs to fix the translations on transifex

blizzz:
rebased

BernhardPosselt:
We should probably add a selenium test case to to reset passwords

mario:
@BernhardPosselt don't have time for this :( Can you do it? I'd like to have this merged and deployed (with the rest of the fixes & improvements I did) as soon as possible.

BernhardPosselt:
Will take a look at it next week end.

rullzer:
Then lets merge this. And do the test in a separate PR so we can deploy this

PR #619, opened 2019-08-14
Fix #251

### LINKED ISSUE #251: Verify email addresses after E-Mail change
When a user changes their email address, it should be verified. allauth provides some views for that which may or may not be useful. Unsure whether email addresses currently are verified at signup, but it would be appropriate for it to use the same mechanism.

mario:


BernhardPosselt:
Good job!

PR #626, opened 2019-11-13
l10n. Removed blanks

rakekniven:
Reported at Transifex. See https://www.transifex.com/nextcloud/nextcloud/translate/#de_DE/$/162523902?q=issue%3Aopen

Signed-off-by: rakekniven <mark.ziegler@rakekniven.de>


rakekniven:
Do we really need the line break here? Is that because of coding rules?

BernhardPosselt:
Yeah, it's the default Python code style linter.

There are many different ways to split lines in Python btw. Maybe this will work better for you:

```python
string = (
    'first part of the line '
    'second part'
)
```

rakekniven:
Better?

PR #646, opened 2020-01-21
Cache DB queries

mario:
Results in caching queries (and invalidating the cache on create/update/delete) to speed things up.

WIP.

### DIFF
--- a/nextcloudappstore/settings/base.py
+++ b/nextcloudappstore/settings/base.py
@@ -16,6 +16,8 @@ from django.conf.global_settings import LANGUAGES
 
 BASE_DIR = realpath(join(dirname(dirname(abspath(__file__))), pardir))
 
+SILENCED_SYSTEM_CHECKS = ["cachalot.E003"]
+
 INSTALLED_APPS = [
     'nextcloudappstore.core.apps.CoreConfig',
     'nextcloudappstore.api.apps.ApiConfig',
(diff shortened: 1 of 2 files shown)


mario:
@BernhardPosselt could you test this pls?

BernhardPosselt:
The django-cachalot is unmaintained and DB level caching is extremely dangerous since it breaks assumptions. Furthermore I'd like to avoid more Django dependencies if possible since they make upgrading hell.

I'd much rather see very direct caching (like last modified timestamps and etags) that work on HTTP level. An obvious target for that is the API which is already cached as good as possible and the web frontend which could make use of that via [template caching](https://docs.djangoproject.com/en/3.0/topics/cache/#the-per-view-cache) or JSON requests if we had a SPA.

mario:
I agree @BernhardPosselt - what I'm looking for are quick wins in terms of performance since it's really needed right now and this seems to provide it while being easily reversible.

Would you accept the following:

`    @method_decorator(cache_page(60*60*2))`

on the ListAPIView resources instead? That would give us a significant boost as well.

BernhardPosselt:
No need to do that, that's already cached with varnish.

BernhardPosselt:
Every relevant API call is cached via Etag afaik

mario:
closing for now.

PR #649, opened 2020-01-29
Added category for featured apps

### LINKED ISSUE #643: add "Featured" category
In order to find "Featured" labeled apps better, there should be a separate category, making all official apps accessible on one screen.

wiswedel:
Signed-off-by: Sascha Wiswedel <sascha.wiswedel@nextcloud.com>

resolves #643

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -354,6 +354,7 @@ category
     * if not provided the category **tools** will be used
     * must contain one of the following values:
 
+       * **featured**
        * **customization**
        * **files**
        * **games**
--- a/nextcloudappstore/api/v1/release/info.xsd
+++ b/nextcloudappstore/api/v1/release/info.xsd
@@ -341,6 +341,7 @@
 
     <xs:simpleType name="category">
         <xs:restriction base="xs:string">
+            <xs:enumeration value="featured"/>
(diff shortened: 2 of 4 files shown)


wiswedel:
Is it enough to just have this category and the featured apps will add themselves automatically, or is there anything else to be done?

juliushaertl:
The main issue with this is that any app developer could add their app there just by adding the category to their apps info.xml file. 

There is already a featured flag that is as far as I know set manually in the django admin page: https://apps.nextcloud.com/?is_featured=true We probably should use that and make the featured listing a bit more prominent.

wiswedel:
> We probably should use that and make the featured listing a bit more prominent.

I agree to use that existing flag and I think it's already prominent enough on apps.nextcloud.com.
However, how do we get a quick-filter for featured apps into the built-in apps page so admins can easily activate them from within the instance?

juliushaertl:
It is available as `isFeatured` in the response of the app store api so we should be able to just create a new list and filter for those in the server.

wiswedel:
> It is available as `isFeatured` in the response of the app store api so we should be able to just create a new list and filter for those in the server.

That's where my competence ends. In any case, I guess this PR can be closed.
Refering back to #643

PR #654, opened 2020-02-04
increase the max download size, as some apps are bigger meanwhile

blizzz:
we patched our  instance like this already…


mario:
Looks good to me.

PR #705, opened 2020-05-04
Fix discourse category creation

mario:
This is a simple copy&paste job for now to fix the fact that discourse category is not created upon integration approval.

Also, since I can't test this - please do it. Thanks!

### DIFF
--- a/nextcloudappstore/scaffolding/forms.py
+++ b/nextcloudappstore/scaffolding/forms.py
@@ -2,6 +2,7 @@ import re
 from os import listdir
 import uuid
 
+import requests
 from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.mail import send_mail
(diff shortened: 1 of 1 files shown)


mario:
Rebased.

nextcloudappstore/scaffolding/forms.py
```diff
@@ -86,6 +87,28 @@ class IntegrationScaffoldingForm(Form):
                                         ' integration '
                                         'does. Can contain Markdown.'))
 
+    def _create_discourse_category(self, app_id: str) -> None:
+        url = '%s/categories?api_key=%s&api_username=%s' % (
+            settings.DISCOURSE_URL.rstrip('/'),
+            settings.DISCOURSE_TOKEN,
```
MorrisJobke, line 93: As of now token and user are `None` ... let me set some up there first :)

mario: Ok. This was used in another place as well...

MorrisJobke: Most likely on an non-authenticated API I think. Do you have a link?

MorrisJobke: Found it ... Anyways - I created one.

mario: :)

PR #758, opened 2020-10-06
Fix invalid tar calls

AJDurant:
The optional --exclude arguments must be before the positional file/folder argument.

#hacktoberfest


bigcat88:
If anything such present for current version fell free to make PR

PR #813, opened 2021-03-22
upgrade webpack and karma

blizzz:
webpack to 4 >= 5.27 (plus deps)
karma to >= 4

locking terser version was released as it is incompatible


blizzz:
when does a github action cancel, and what is actually happening? 

Screenshot_20210322_170328

blizzz:
output looks good, some deprecation messages, but otherwise ok.

Issue #834, opened 2021-04-23
Generated applications have an invalid name in Composer 2.0


srbaker:
When generating an application from the store, it recommends using a CamelCase name, and puts that as `name` in the `composer.json`

Since Composer 2.0, the `name` field must be all lowercase and numbers, and must contain a vendor part i.e.  `foo/bar` or `nextcloud/tasks`.  The validity regex is `^[a-z0-9]([_.-]?[a-z0-9]+)*/[a-z0-9](([_.]?|-{0,2})[a-z0-9]+)*$`

(See https://getcomposer.org/doc/04-schema.md#name)

Following the directions for my project FocusFrog generates a `composer.json` containing `"name": "FocusFrog"`.

I am happy to fix this myself and contribute it, but we need to add a field to the generator that represents "vendor" and I'm not sure how this should be presented to the user.

bigcat88:
that is already fixed

PR #836, opened 2021-04-23
[WIP] Update the app generation scaffold.

srbaker:
I noticed a few issues when generating a new application.

Composer 2.0 didn't work (reported in issue #834), and the tests wouldn't run because the default Integration test was for a new version of PHPUnit, but the default `composer.json` set it to some version around 5.

I've updated PHPUnit to 9 (8 was EOL in February), and have fixed the tests to work with PHPUnit 9.

More changes will be forthcoming, but this is WIP because I have some issues I'd like help with:

1. How do we deal with Composer 2.0+'s new requirement that `name` is in the format `vendor/project-name` and requires that both be lowercase? (Mentioned in #834)
2. What version _should_ this apply to? I put 22, because that's next-release, but I'm sure this should be backported to whenever it stopped working.
3. I didn't see any tests that generate an app, and then run the tests. That would catch these kinds of thing, I intend to write tests for this unless someone can provide a good reason not to.

My goal is that when a user generates an app, they should be able to run `make test` successfully, without error.  This is the fastest way (for me, at least) to get started.

Any other thoughts/concerns, please put them in the comments. I want to encourage people to make nextcloud apps, and the generator is the starting point, so it's really important to me that it works well.

Thanks!


blizzz:
Maybe we should just provide https://github.com/nextcloud/app-tutorial 

It is too bloated for a pure scaffold, on the other hand it is maintained and up to date.

CarlSchwan:
> Maybe we should just provide https://github.com/nextcloud/app-tutorial
> 
> It is too bloated for a pure scaffold, on the other hand it is maintained and up to date.

Just did so in https://github.com/nextcloud/appstore/pull/915

PR #921, opened 2022-08-24
Handle malformated tar files

CarlSchwan:
When getting the list of names in the tar files, we should only get valid members of the tar file. This is not always the case since recent version of python and we need to handle the exception.


miaulalala (approved):

blizzz (approved):

blizzz:
we should get the tests sorted out at some point…

PR #924, opened 2022-09-02
Fix login issues

CarlSchwan:
- Don't send email when only the hash of the password changed due to an
  upgrade of the hashing algorythm used
- Wrap deleting and recreating token in a transation to prevent data
  race

Signed-off-by: Carl Schwan <carl@carlschwan.eu>


miaulalala (approved):

juliushaertl (approved):
Code makes sense

AndyScherzinger (approved):

cmweathe (commented):
[newbie]
In progress

PR #933, opened 2022-11-02
Fix CSP when logging with github on chromium

CarlSchwan:
Chromium requires the redirected url to also be in the csp rules. This is an inconsistent behavior between firefox and chrome


CarlSchwan:
(already patched the server so this is why it works now)

blizzz (approved):

PR #938, opened 2022-11-23
Improve unstable detection

nickvergessen:
1.25.2-1 is not an unstable version
Ref https://apps.nextcloud.com/apps/talk_matterbridge

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -419,7 +419,10 @@ class AppRelease(TranslatableModel):
 
     @property
     def is_unstable(self):
-        return self.is_nightly or '-' in self.version
+        return (self.is_nightly or '-dev' in self.version
+          or '-a' in self.version or '-alpha' in self.version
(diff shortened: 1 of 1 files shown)


SystemKeeper:
Maybe check case insensitive ?

nickvergessen:
> Maybe check case insensitive ?

Not needed as php version_compare does not check case insensitive either. Now the handling reflects what the server does.
https://www.php.net/manual/de/function.version-compare.php

blizzz (approved):

blizzz (approved):
:snake:

SystemKeeper:
Okay, makes sense to align that. I have no 
experience in using the NC App Store. Is this documented somewhere that this is aligned with the php Check?

nickvergessen:
https://nextcloudappstore.readthedocs.io/en/latest/developer.html#info-xml states semantic version

CarlSchwan (approved):

nickvergessen:
Can we get this deployed?
cc @blizzz

PR #948, opened 2022-12-25
Fix issue with double curly braces in GitHub actions

christianlupus:
Currently, some of the GitHub workflows are not working as they are parsed by Django and thus the double curly braces are broken. This PR should just fix these cases


AndyScherzinger:
@juliushaertl what do you think?

juliushaertl:
Makes sense in general. I tried this locally but for me all the hidden files were actually missing from the tar ball. Didn't have time to dig into why. It seemed unrelated to this PR but I wanted to check that again before approving.

christianlupus:
Just ping me if you want me to rebase and incorporate the latest changes in master.

bigcat88 (approved):

bigcat88:
yeah, this worked for me locally, thanks.

PR #952, opened 2023-01-09
feat(scaffold): Implement optional csv logs for app scaffold requests

juliushaertl:
If a path is specified in the config like `APP_SCAFFOLDING_LOG = '/tmp/newapps.csv'` every request for an app scaffold can be written to a csv file

Example CSV:
```
MySmallNewApp|21|Julius Härtl|admin@admin.com|http://homepage.com|https://github.com|['dashboard', 'multimedia', 'search']|My short summary|A longer description
```

- [ ] Check ci failures (might be on master as well)
- [ ] Check back with @DaphneMuller if that is sufficient
- [ ] Add note to the frontend that we collect information here

### DIFF
--- a/nextcloudappstore/scaffolding/views.py
+++ b/nextcloudappstore/scaffolding/views.py
@@ -11,6 +11,7 @@ from nextcloudappstore.scaffolding.forms import AppScaffoldingForm, \
 from django.conf import settings
 
 import csv
+from datetime import datetime
 
 
 class AppScaffoldingView(FormView):
@@ -31,8 +32,9 @@ class AppScaffoldingView(FormView):
     def form_valid(self, form):
         if settings.APP_SCAFFOLDING_LOG:
             with open(settings.APP_SCAFFOLDING_LOG, 'a') as csvfile:
+                entries = form.cleaned_data.values()
                 logwriter = csv.writer(csvfile, delimiter='|')
-                logwriter.writerow(form.cleaned_data.values())
+                logwriter.writerow([datetime.now(), *entries])
         buffer = build_archive(form.cleaned_data)
         response = HttpResponse(content_type='application/tar+gzip')
         response['Content-Disposition'] = 'attachment; filename="app.tar.gz"'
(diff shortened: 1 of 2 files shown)


nextcloudappstore/scaffolding/views.py
```diff
@@ -25,6 +28,10 @@ def get_initial(self):
         return init
 
     def form_valid(self, form):
+        if settings.APP_SCAFFOLDING_LOG:
+            with open(settings.APP_SCAFFOLDING_LOG, 'a', newline='') as csvfile:
```
CarlSchwan, line 32: Since this contains personal information, a notice should be added in the webpage.

Aside from that code looks good 👍

juliushaertl: Yep, good point.

DaphneMuller:
this looks good @juliushaertl , does the entry include a date stamp?

juliushaertl:
Not yet, but I was about to ask if there is more things you may want to have. Will add the date and time stamp as well :)

juliushaertl:
> awesome!! then it's complete. Only question rests is how I will be able to access the file

@AndyScherzinger suggested that sysadmin could automate that then to either send it to you by mail or figure out a different way.

AndyScherzinger:
@juliushaertl does it write a single file or 1 per item? And what is the path?

juliushaertl:
A single file, the path can be specified through the environment settings by adding the following to production config for example `nextcloudappstore/settings/production.py`:

```
APP_SCAFFOLDING_LOG = '/tmp/newapps.csv'
```

AndyScherzinger:
So I guess "just purging its content now and then" shouldn't cause any trouble?

juliushaertl:
There is no locking of course so there is some risk when exacting the file content and purging it that inbetween writes would be lost, but that is likely a timeframe of some milliseconds. We can get more sophisticated, but this seemed enough for the use case. Appending to the file is safe for concurrency.

juliushaertl:
And the file will be recreated (assuming permissions are granted to the process running the app store) if it was purged before.

AndyScherzinger:
Sounds good and good enough from my perspective 👍

PR #953, opened 2023-01-09
style: Fix pylint errors

juliushaertl:
Let's see what ci thinks now :)


juliushaertl:
Merging since it is a ci change only and passes now.

PR #957, opened 2023-01-11
Add script to template in order to bump app versions

christianlupus:
This set of scripts should allow the developer to bump the version of the app to a new version. It will update the `appinfo/info.xml` and the `package.json` file with the first CLI parameter.
By adding custom scripts to the appropriate folder, it is possible to update other files automatically as well.

There is a [topic in the forum](https://help.nextcloud.com/t/update-of-scripts-in-the-app-template/153411) that poses some general questions that might be correlated with this PR.


bigcat88:
I suggest to implement this in GitHub Actions, when owner of app push tag - replace with commit automatically those versions.

christianlupus:
Adding a github action could be done as an add-on. This structure was designed to allow custom scripts to be added easily without much effort. Running these scripts from a GH action is no big deal. This would allow the authors to disable the GH action though (makes not much sense, I have to admit).

I can provide an implementation (after making it mergable) for that if that would be accepted and merged. Though, I do not want to *work for nothing* (aka the PR be rejected in the end). Therefore I am asking before investing too much and on too many details.

bigcat88:
If you could add entry for this to the Makefile(for example: `make set-version 2.2.3`), I will gladly merge such a pull request.

Name of makefile command can be any that you choose.

christianlupus:
Make dows not allow for arbitrary parameters. (All parameters are interpreted as Make targets.)

I could pass it via env variable (that would be like `make update-version VERSION=1.2.3`) or try to create a template target like `make update-version-1.2.3`. The latter might need a clear structure (only 3 parts) or a gernic match (so `make update-version-new` would work as well to create a version `new`).

I personally prefer the env var variant as it seems cleaner. Do you consent, @bigcat88?

bigcat88:
> Make dows not allow for arbitrary parameters. (All parameters are interpreted as Make targets.)
> 
> I could pass it via env variable (that would be like `make update-version VERSION=1.2.3`) or try to create a template target like `make update-version-1.2.3`. The latter might need a clear structure (only 3 parts) or a gernic match (so `make update-version-new` would work as well to create a version `new`).
> 
> I personally prefer the env var variant as it seems cleaner. Do you consent, @bigcat88?

Yep, I am totally fine with this, if it will work through env like you suggested.

bigcat88:
Let know when this is ready to merge

christianlupus:
With setting the app template into a dedicated submodule, this is most probably no longer needed. I close the PR therefore here. If you think it is still useful, please give feedback!

PR #958, opened 2023-01-16
fix(scaffold): Make storing personal data opt in for app scaffording

juliushaertl:
Changing the checkbox to opt in for personal data storing. Otherwise we store only the timestamp and app name.


DaphneMuller:
sounds good @juliushaertl thank you 🤗

CarlSchwan (approved):
🥰

Issue #990, opened 2023-04-28
Add Nextcloud 26 and 27 to the app skeleton generator


DaphneMuller:
Add Nextcloud 26 and 27 to the app skeleton generator in the Nextcloud version field

## Steps to reproduce

1. Go to https://apps.nextcloud.com/developer/apps/generate
2. Select 'nextcloud version'
Expected result: User can choose Nextcloud 26 or maybe even already Nextcloud 27
Actual result: User can only choose until Nextcloud 25
<img width="1206" alt="Screenshot 2023-04-28 at 13 45 25" src="https://user-images.githubusercontent.com/86835268/235139205-f30c307a-b1d4-47f6-be15-04ec8fdac99a.png">

PR #997, opened 2023-05-05
templates for NC 26 and 27

bigcat88:
Resolves: #990 

@andrey18106 can you take a look what I had missed in skeleton templates and update if missed something(maybe some lock files needs to be more fresh and etc).
If will found something commit fixes to this PR please.

CI will fail until we merge #996, so do not merge it before fix for CI will arrive


bigcat88:
I'll rewrite this

PR #1010, opened 2023-05-05
Updated poetry.lock

bigcat88:
1. Updated python dependencies(3 critical security and 3 high security alerts)
2. Added timeouts for requests(big timeouts with 30 and 21 second so should not break anything)
3. Replaced all deprecated of methods and functions


mejo- (approved):
:elephant:

PR #1017, opened 2023-05-07
app-templates maintenance

bigcat88:
Fixes: #990 

1. Little updated actions for generated templates(but in future we should take them from https://github.com/nextcloud/app-tutorial)
2. Removed 21,22, 23 templates(as when this AppStore update will be deployed NC 27 will be in, so range 24-27 is fine)
3. Added 26,27 templates as symlinks to 25 version


AndyScherzinger (approved):
🐘

PR #1051, opened 2023-05-29
Fix the baseline file in the boilerplate to be valid XML

christianlupus:
The current boilerplate code has an invalid XML file included that breaks the psalm tests (the baseline file cannot be read correctly and thus psalm refuses to work).

This is just a quick fix to have a working psalm configuration.


bigcat88 (approved):

bigcat88:
@christianlupus thanks!

PR #1056, opened 2023-06-02
switching from coveralls to codecov.io

bigcat88:
_old coveralls action use old docker image thus does not support self-hosted well_


bigcat88:
_Required check for coverage was removed, after this pull request will be merged, will add the required check for codecov action instead_

andrey18106 (approved):

PR #1163, opened 2023-10-04
dropped python3.9, added ruff

bigcat88:
We do not need Python 3.9 at all, as Ubuntu 22 has Python 3.10 by default.

P.S.: In next PR will add `ruff` checker with basic rules, to speed up writing the correct formatted code.


bigcat88:
Will add more `ruff` checks in this PR and completly remove `flake8`

bigcat88:
Done: Minimum satisfactory number of checks possible

andrey18106 (approved):

PR #1218, opened 2023-11-12
update app template

kesselb:
Howdy :wave: 

I recently used the generator and ran into a few smaller things:

1) The file name for the default php-cs-fixer was wrong
2) Every app should use IBoostrap nowdays
3) christophwurst/nextcloud is nextcloud/ocp for a while

Note: I don't have a working dev setup for the appstore and therefore couldn't test the app skeleton generator. 
Could one provide my a copy of the generated skeleton app to double-check my changes?


kesselb:
cc @ChristophWurst okay to always add IBootstrap when using the generator?

bigcat88:
[app-2.tgz](https://github.com/nextcloud/appstore/files/13331708/app-2.tgz)


> Note: I don't have a working dev setup for the appstore and therefore couldn't test the app skeleton generator. Could one provide my a copy of the generated skeleton app to double-check my changes?

ChristophWurst (approved):
Indentation is off, though

nextcloudappstore/scaffolding/app-templates/25/app/lib/AppInfo/Application.php
```diff
@@ -6,11 +6,30 @@
...
 
-class Application extends App {
+class Application extends App implements IBootstrap {
 	public const APP_ID = '{{ app.id }}';
 
 	public function __construct() {
 		parent::__construct(self::APP_ID);
 	}
+
+    public function register(IRegistrationContext $context): void {
```
bigcat88, line 20: yes, as Christophe mentioned, keeping indentation consistent will be helpful here.

kesselb:
Thanks :+1: 

The "generated" version is really helpful. I've pushed another commit to also update platform.config to php8. I'm aware that's not 100% correct because the app template for 25 should still support php7.4, but 25 and php7.4 are both EOL, and otherwise you cannot run composer install because nextcloud/ocp needs at least php8.0.

You are right about the indentation. It's unfortunately not in line with our code style with a couple of files.  I've run composer run cs:fix to update the files.

kesselb:
I'm afraid neither docker-compose up nor make dev-setup works for me. Both fail with some error.

@bigcat88 could you please generate another one?

bigcat88:
[app-3.tgz](https://github.com/nextcloud/appstore/files/13335488/app-3.tgz)

here it is.

> because the app template for 25 should

We can probably drop the templates for 25 NC and set minimum 26 version?

bigcat88:
Is this PR no longer needed?

kesselb:
> Is this PR no longer needed?

No. Thanks for the reminder.

PR #1234, opened 2023-12-05
add app-template for 26

kesselb:
https://github.com/nextcloud/appstore/pull/1218 went stale due to many changes ...

Let's tackle it in smaller steps ;)

- Added a dedicated app-template for 26 instead of the symlink to 25
- Application class is now using IBootstrap
- nextcloud/ocp for the matching nextcloud version is installed
- Code style applied


kesselb:
@bigcat88 I still don't have a working dev setup, would you mind generating another app with this branch and sharing the result?

bigcat88:
[app.tar.gz](https://github.com/nextcloud/appstore/files/13560124/app.tar.gz)

kesselb:
> [app.tar.gz](https://github.com/nextcloud/appstore/files/13560124/app.tar.gz)

Thank you very much :+1: 

- The generated version for nextcloud/ocp seems right
- composer install works
- composer run cs:fix works and does not find any error (depending on the app name it may find some, but that's unavoidable) 
- npm install works
- npm run dev works

This looks okayish to me. At least it's a bit better than before ;)

ChristophWurst (approved):

Issue #1235, opened 2023-12-07
App upload to the appstore sometimes fails to store the proper app info to database


MB-Finski:
## Details

As a continuation of the discussion from here: https://github.com/nextcloud/integration_openai/issues/64

Upon update of an app, somewhat randomly, the description, name and summary field are not parsed/stored properly such that the apps.json returned by https://apps.nextcloud.com/api/v1/apps.json does not contain the necessary items under the "translations" key for the app in question. Due to a subsequent parsing error in the Nextcloud frontend AppList.vue the apps page under settings stops loading for all Nextcloud users.

A snippet from AppList.vue responsible for the parsing error:
> apps() {
>   const apps = this.$store.getters.getAllApps
>   .filter(app => app.name.toLowerCase().search(this.search.toLowerCase()) !== -1)
>   .sort(function(a, b) {
> 	  const sortStringA = '' + (a.active ? 0 : 1) + (a.update ? 0 : 1) + a.name
> 	  const sortStringB = '' + (b.active ? 0 : 1) + (b.update ? 0 : 1) + b.name
> 	  return OC.Util.naturalSortCompare(sortStringA, sortStringB)
>   })
> 

The synopsis of the issue by @bigcat88 from the integration_openai issue:

> My observations (not necessarily true):
> 
> - The problem was not in the translations, just for some reason, when uploading the release to the app store, new data was not uploaded to the fields: name, summary, description.
> - No errors were found in the database logs.
> - Errors in the app store logs clearly indicating why the entry to these fields did not occur either.
> - The code that is responsible for this was written a long time ago and has not changed; in a couple of hours I did not find any errors in it.
> 
> My conclusion: some kind of memory error occurred, which happens once in a million cases, or some kind of race-condition, that is not related to this project(integration_openai).
>

bigcat88:
New observations:

_Under debugger if I put breakpoint on 301 line with "translation.delete()" I always get bug when old translations got deleted and new does not inserted. Looks like some bug with debugger or underlying "parler" / "psycopg" library._

As I understood the algorithm from the source code, all translations were deleted so that if a new release was not uploaded in the current language of the application, there would be no old translations left.

Linked PR behave the same way, but if the new release language is the same, it skips the  deleting, and record will be just updated in database and this should prevent this bug.

bigcat88:
in last two weeks all is fine with this, so closing it for now.

PR #1237, opened 2023-12-08
clear only translations that are not present in the new release

bigcat88:
Relates to #1235

Do not merge this, until we will be sure that it helps, we will wait for one week and see if it helps.


andrey18106 (approved):

bigcat88:
looks like this fixes it, somehow

PR #1247, opened 2023-12-15
Update README.md

NithurshNITHURSHA:


bigcat88:
Current Version:

"We do not handle Nextcloud integration." **This sentence is direct and emphasizes what our team does not do.**

It's clear and straightforward, making it suitable if we want to emphasize the limitation of our service right away.

Proposed Changes:

"Nextcloud integration is not handled by us." 

It shifts the focus slightly, putting the emphasis on "Nextcloud integration" rather than on "we/us."

Your proposed version is also correct and might be perceived as slightly more professional or formal, but I do not see why we should change existing one.

bigcat88 (changes requested):
_previous version is correct too_

Issue #1221, opened 2023-11-22
Report / Flag Review Spam


meichthys:
Please provide the following details for error reports from https://apps.nextcloud.com/. Ignore in case of a feature request.

## Details

I've seen a number of reviews on apps that are obviously spam. It would be nice if there was a way to report/flag a review as potential spam - to be later reviewed by the Nextcloud team.

One example of potential spam is the review on this app: https://apps.nextcloud.com/apps/memories

User: `pdo' pdowski`
Review content: `"><img%20src="https://webhook.site/700c721c-e6be-47ae-8fad-2db54c17ccf9">`

Fred-Zweig:
You are right. Many reviews are abusive or spam. Other developers signaled this problem [here](https://github.com/nextcloud/appstore/issues/1215) and [here](https://github.com/nextcloud/appstore/issues/947). We hope that someone from the Nextcloud team will implement a way to report such reviews, so that later they can be checked by an admin and removed.

bigcat88:
@andrey18106 can we easy implement this?

From frontend part we need a button on each comment only for the App author, that will set new field(`appeal`) in `AppRating` model to 1/True

And for Admins we implements the same page as "apps.nextcloud.com/account/integrations" but for reviewing marked Ratings.

Should not be very difficult, imho. What do you think?

alexanderdd:
Hm, I am logged in (regular user) but I don't see a "report"-button for ratings? 

For example, in English this page https://apps.nextcloud.com/apps/documentserver_community has a comment "**** THE RELEASERS!" which I wanted to report.

bigcat88:
> Hm, I am logged in (regular user) but I don't see a "report"-button for ratings?

currently **only** authors of apps can report comments

PR #1256, opened 2023-12-24
Comments: Report / Flag Review Spam 

bigcat88:
This PR implements flagging of comments for an "appeal" that the Administrators user group can accept and the comment/rating will be deleted or rejected and left as is.

Closes: #1221


bigcat88:
@andrey18106  I added migration to the existing new one, so before testing delete last migration and apply from scratch

andrey18106 (approved):

PR #1265, opened 2024-01-02
apprating-cache bug

bigcat88:
Ref: https://github.com/nextcloud/appstore/pull/1256

Here we forgot to update caches, so this PR fixes it.

Added new field `last_modified` and ETag is calculated based on it instead of `rated_at` which now is the date of AppRating creation only.


andrey18106 (approved):
LG

PR #1355, opened 2024-04-17
Remove app-templates for EOL versions

provokateurin:
I split it up because the rename detection was going a bit wrong (ok Github still shows it wrong if you look at the overall diff).


bigcat88:
the template from 26 version should not be touched, since the templates for 27 and 28 are symlinks for it

provokateurin:
I didn't touch it, I just replace the 27 template with the one for 26

nextcloudappstore/scaffolding/tests/data/info.xml
```diff
@@ -16,7 +16,7 @@
     <category>tools</category>
     <bugs>https://test.com</bugs>
     <dependencies>
-        <nextcloud min-version="24" max-version="24"/>
+        <nextcloud min-version="28" max-version="28"/>
```
provokateurin, line 19: Shouldn't the min version be 27 because the template is also for 27?

bigcat88: tests were modified to test against 28 version, to not modify them in 3 months again..

bigcat88 (approved):

PR #1368, opened 2024-04-29
refactor(scaffolding): Migrate app template to Github template repo

provokateurin:
As an alternative to downloading the zip on the fly the app_template repo could be added as a git submodule to pin a specific version to avoid unintended changes. This submodule would need to be updated at least once per server release which adds more work again (quite manageable though).


bigcat88:
Can we just redirect the user when clicking to start an automatic download via a redirect, will it be fine?

This will save us from adding caching of this, because in the current implementation, sending a download request and not caching the response will lead to a fairly simple opportunity to DDOS AppStore in some cases.

provokateurin:
You are right, the problem is that it's not simply downloading and returning the zip file, but it replaces the content in some files so the values the user submitted are applied.

bigcat88 (approved):

PR #1419, opened 2024-07-23
feat: add warning banner to apps not updated within the last 3 major releases of Nextcloud

### LINKED ISSUE #1395: Add banner to an app's page if hasn't been updated to run with any of the last 3 major releases of Server
Similar to how WordPress.org handles old plugins that seem no longer maintained:

> This plugin hasn’t been tested with the latest 3 major releases of WordPress. It may no longer be maintained or supported and may have compatibility issues when used with more recent versions of WordPress.

Gets displayed immediately under the screenshot:

e.g. https://wordpress.org/plugins/flush-opcache/

edward-ly:
Closes #1395.

Screenshot 2024-07-23 at 13-33-22 App Store - Nextcloud


nextcloudappstore/core/models.py
```diff
@@ -254,6 +254,19 @@ def compatible_unstable_releases(self, platform_version, inclusive=True):
...
+    def is_outdated(self):
+        """Checks if an app has been released in last 3 recent platform versions
+
+        :return: True if not compatible, otherwise false
+        """
+
+        release_versions = list(self.latest_releases_by_platform_v().keys())
+        if not release_versions:
+            return True
+        max_release_version = Version(pad_max_inc_version(release_versions[0]))
```
bigcat88, line 266: Unfortunately, I don't think this will work as `release_versions` is not sorted.
Have you tested this?

edward-ly: They should be sorted, no? Look in the implementation for `latest_releases_by_platform_v`.

bigcat88: Can you point out the place in the implementation where it sorts them?

edward-ly: Ah sorry, I must have confused it with another function somewhere along the way. Will fix it now.

bigcat88 (approved):
With the latest changes it looks good

PR #1426, opened 2024-07-27
feat: add relevance sort to sorting options

### LINKED ISSUE #1298: improve search results order
Hello everyone!

IMO the search results order should be improved. More relevant apps should be shown first.

For example, search for "mail" - the most relevant app is "Mail" which is shown in fourth row, I have to scroll to see it. Second relevant would be Snappy Mail. The other results should be below that. Rainloop is deprecated and should not be shown at all - instead, it is shown as the first result.
Screenshot 2024-02-11 at 23-30-35 Alle Apps - App Store - Nextcloud

Or search for "map" - the most relevant app is "Maps", shown in third row, I have to scroll to see it.
Screenshot 2024-02-11 at 23-33-43 Alle Apps - App Store - Nextcloud

Thanks for reading =)

edward-ly:
Closes #1298, replacing rating as the default sort with and without a search query. At the moment, it's only partially tested; I may refine results a bit more depending on the various search results I get.

Screenshot 2024-07-26 at 17-30-03 All apps - App Store - Nextcloud


bigcat88:
```
http://127.0.0.1:8000/?order_by=relevance&ordering=desc&search=app_api
4.2.14
NotSupportedError
Calling QuerySet.prefetch_related() after union() is not supported.
/Users/user/PycharmProjects/appstore/.venv/lib/python3.11/site-packages/django/db/models/query.py, line 1960, in _not_support_combined_queries
nextcloudappstore.core.views.CategoryAppListView
/Users/user/PycharmProjects/appstore/.venv/bin/python
```

edward-ly:
I finally figured out a workaround to the `union` operator, so the sort should be working now.

bigcat88 (approved):
seems to work without errors - everything is great

PR #1452, opened 2024-08-15
feat: add donation attribute to `appinfo/info.xml`, donation buttons to app details page

edward-ly:
One of two requirements for a new payment system on the app store (see [here](https://cloud.nextcloud.com/f/10131205) for details).

Screenshot 2024-08-14 at 22-55-38 App Developer Guide — AppStore 4 3 3 documentation

Screen Shot 2024-08-14 at 23 03 27


bigcat88 (approved):
Slightly tested,- overall looks very solid and nice.

docs/developer.rst
```diff
@@ -211,6 +211,8 @@ A full blown example would look like this (needs to be utf-8 encoded):
         <repository>https://github.com/nextcloud/news</repository>
         <screenshot small-thumbnail="https://example.com/1-small.png">https://example.com/1.png</screenshot>
         <screenshot>https://example.com/2.jpg</screenshot>
+        <donation type="paypal" title="Donate via PayPal">https://paypal.com/example-link</donation>
```
susnux, line 214: Would it not be more consistent to wrap it like we do with all other lists?
E.g. types, dependencies etc

So like
```xml
<donations>
    <donation title="..." href="..." />
    <donation .... />
</donations>
```

edward-ly: I'm not sure if I understand what added benefit that would serve. Plus, there are plenty other lists that aren't wrapped in the same way as well (author, category, screenshot, etc.).

susnux: > category, screenshot

Yes I know :sweat_smile: 
So for me it is not important, just that those files are written by hand and then it is often easier to group everything. Also the DOM parsing is a bit easier.

But as said, not really important :)

PR #1467, opened 2024-09-04
optimized viewing of the AppStore pages

bigcat88:
Measurements were made locally on a fairly large database with records.

1. It was found that one of the latest changes simply incredibly loads the AppStore just when you view pages:

```python3
   context["is_outdated"] = False  # self.object.is_outdated() 
```

**Only this change is already present in the production, I made it yesterday.**

**Impact of it ultra huge**, for one page view it increases open time of page from `2s` to `3s` locally in my setup.
In AppStore yesterday it was like more `5s` in difference with enabled and disabled.

@edward-ly  can you take a look to thing how we can cache this or reimplement?

---

2. The remaining optimizations were applied to the original AppStore code, which was always like this, but since there were fewer applications and their releases in it before, this problem did not arise.

I did not apply these changes on the server, I only tested them locally.

Although they are small, these lines that are changed here are called dozens of times, for each version for each application, so the impact of their changes is significant.

```python3
   from semantic_version import SimpleSpec, Version
```

Impact: 3%-8%

``Spec`` is deprecated in raising invisible warnings inside each call. Replaced with non-deprecated version.

```python3
    filter(lambda r: (not r.is_unstable) and r.is_compatible(platform_version, inclusive), self.releases.all())
```

Impact: 5%-10%

Explanation: first faster to check for `bool` value, and after that starting parsing with custom functions.

```python3
   filter(lambda r: r.is_unstable and r.is_compatible(platform_version, inclusive), self.releases.all())
```

Impact: 5%-10%

Explanation: first faster to check for `bool` value, and after that starting parsing with custom functions.

```python3
    return min_version in spec or Version(pad_max_inc_version(platform_version)) in spec
```

Impact: 3%-5%

Explanation: Here we do not need to init with padding version and it should be in the right side to be per foment only if first condition does not triggered.

Overall, a rollback of the change introduced by Edward to determine whether a release is outdated or not + these 4 small changes in parsing each release - I hope will help reduce the load on the AppStore.

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -30,7 +30,7 @@ from packaging.version import InvalidVersion
 from packaging.version import parse as parse_version
 from parler.models import TranslatableManager  # type: ignore
 from parler.models import TranslatableModel, TranslatedFields
-from semantic_version import Spec, Version
+from semantic_version import SimpleSpec, Version
 
 from nextcloudappstore.core.facades import distinct
 from nextcloudappstore.core.rating import compute_rating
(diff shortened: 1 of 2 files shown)


SystemKeeper (approved):

bigcat88:
> @edward-ly can you take a look to thing how we can cache this or reimplement?

never mind, in this PR #1468 caching is added to the low-level `releases_by_platform_v` and `unstable_releases_by_platform_v` methods and **"outdated release"** feature can be enabled again from what I see.

PR #1470, opened 2024-09-05
consider "integrations" always up to date

bigcat88:
Integrations does not have releases and was always marked as "outdated".

Reported by Simon, he spotted it on the AIO integration marked as outdated.

This fixes this(already deployed for test)


AndyScherzinger (approved):

szaimen (approved):
Seems to work. Thanks for the fast fix :)

PR #1492, opened 2024-09-19
fix: correctly import donation tag attributes

edward-ly:
See my [Zulip integration app](https://apps.nextcloud.com/apps/integration_zulip) for an example. This is what is displayed right now vs. what it should look like with this PR.

| Before | After |
| --- | --- |
| Screenshot 2024-09-19 at 11-22-26 Zulip Integration - Apps - App Store - Nextcloud | Screenshot 2024-09-19 at 11-22-44 Zulip Integration - Apps - App Store - Nextcloud |


bigcat88 (approved):
LGTM

PR #1496, opened 2024-09-26
fix(App): change default discussion URL to one that is always valid

### LINKED ISSUE #1279: generate valid default app discussion URL by default
## majority of the apps in the store don't have valid "discussion" URL
Follow up of  #1246. While the previous bug was addressing one specific app this is about the whole app store. if the dev doesn't provide a discussion URL  in the [app metadata](https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html) the store generates default URL pointing to non-existing category in help.nextcloud.com forum. 

As result:
- from 266 apps compatible with NC28 
- only 26 apps have valid URL 
- 13 of the 26 use custom URL outside of the forum

## root cause
Apps can provide custom discussion URL using info.xml app metadata. But the "discussion" value is optional and in case the value is not populated the store generates a discussion URL pointing to the `https://help.nextcloud.com/c/apps/` + %appid% but there is no process to create the category in the forum. additionally the forum team is going to get rid of the categories and moving to more modern "tags" approach.

from [App metadata — Nextcloud latest Developer Manual latest documentation](https://docs.nextcloud.com/server/latest/developer_manual/app_development/info.html)

    discussion
    - optional
    - must contain a URL to the project’s discussion page/forum
    - will be rendered on the app detail page as the “ask question or discuss” button
    - if absent, it will default to our forum at https://help.nextcloud.com/ and create a new category in the apps category

## Solution
To address the situation there are two option
- adopt the default discussion URL to modern forum approach and target app-specific "_forum tag_" e.g. `https://help.nextcloud.com/tags/%appid%-app`
- change the default discussion URL to the forum main page `https://help.nextcloud.com`
my personal preference is the second as it seems to be much leaner and allows restructuring the forum without app store changes.

edward-ly:
Resolves #1279, changing the default discussion URL to https://help.nextcloud.com/.


bigcat88:
will this work for Talk(spreed app)?

edward-ly:
> will this work for Talk(spreed app)?

https://github.com/nextcloud/spreed/blob/main/appinfo/info.xml
Looks like no, but I would suggest adding a `<discussion>` tag with the link in this case.

bigcat88 (approved):
anyway, we can deploy this and after add exceptions for all apps that breaks

bigcat88:
> > will this work for Talk(spreed app)?
> 
> https://github.com/nextcloud/spreed/blob/main/appinfo/info.xml Looks like no, but I would suggest adding a `<discussion>` tag with the link in this case.

feel free to make a PR for that in "spreed" =)

SystemKeeper:
Just so you are aware, we are currently migrating from categories to tags on help.nextcloud.com.

SystemKeeper:
> So merging and deploying it, will not harm :)

No, that should be fine. The linked PR still referenced a category, that's why I mentioned that the migration is ongoing and we should not add new links to categories anymore.

tobiasKaminsky:
Then please deploy, or do you need help from sysadmin team on this?

PR #1497, opened 2024-09-27
allow "scopes" be not present in "external-app"

bigcat88:
we decided to abandon AppAPI scopes in AppAPI and the last three versions of AppAPI have already been released where it completely ignores them.

it's time to add the ability to not specify them in `info.xml` at all, since they are no longer needed.

we are not cutting them out of the AppStore completely yet, we need to wait at least six months so that those who have the old version of AppAPI now do not get crashes (old versions of AppAPI need this data and they expect that they will be)


edward-ly (approved):
LGTM

PR #1518, opened 2024-12-06
implemented Odoo integration(newsletter mail list)

oleksandr-nc:
**Task from Viktor Nagornyy**

Integration with the Odoo server has been implemented; a checkbox is available to the user upon registration:

image

And also in the settings of each account:

image

**By default it is disabled.**

When the status of this checkbox changes before a DB request, network requests are sent to Odoo when a subscription is enabled or disabled.

---
**Deploy notes:**

```
# Fill these values in the "production.py" file:
ODOO_URL = ""
ODOO_DB = ""
ODOO_USERNAME = ""
ODOO_PASSWORD = ""
ODOO_MAILING_LIST_ID = 0
```

This PR adds new table, so applying migrations is necessary, and also recompile the frontend(`npm run build`) to take effect.

Also the command `python manage.py create_missing_profiles --settings nextcloudappstore.settings.production` should be run to create entries in the newly created table.

*PS: I did not tested when registering using GitHub, I can't do that easy in dev-setup, so it will be tested after deploy.*


nextcloudappstore/user/forms.py
```diff
@@ -11,19 +11,33 @@
...
+
 
 class SignupFormRecaptcha(forms.Form):
     """integrate a recaptcha field."""
 
     captcha = CaptchaField()
     first_name = CharField(max_length=30, label=_("First name"))
     last_name = CharField(max_length=30, label=_("Last name"))
+    subscribe_to_news = forms.BooleanField(
+        label=_("I agree to receive app developer news and updates from Nextcloud by email (optional)"),
```
edward-ly, line 24: Just a small point, but the word "agree" doesn't feel like the right word to use for me as it suggests some kind of contractual agreement that is required when in this case it's actually not. `I would like to receive app developer news and updates from Nextcloud by email (optional)` sounds better, I think.

nextcloudappstore/user/migrations/0001_initial.py
```diff
@@ -0,0 +1,33 @@
...
+    ]
+
+    operations = [
+        migrations.CreateModel(
+            name="UserProfile",
+            fields=[
+                ("id", models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name="ID")),
+                (
+                    "subscribe_to_news",
+                    models.BooleanField(default=True, help_text="User agrees to receive Nextcloud news and updates."),
```
edward-ly, line 23: Similarly, `User has opted in to receive Nextcloud news and updates.` might be a better sentence here.

nextcloudappstore/user/forms.py
bigcat88: thanks, that makes sense

nextcloudappstore/user/forms.py
```diff
@@ -65,10 +79,24 @@ class AccountForm(forms.ModelForm):
             "password!"
         ),
     )
+    subscribe_to_news = forms.BooleanField(
+        label=_("I agree to receive app developer news and updates from Nextcloud by email (optional)"), required=False
```
edward-ly, line 83:
```suggestion
        label=_("I would like to receive app developer news and updates from Nextcloud by email (optional)"), required=False
```
same here

nextcloudappstore/user/models.py
```diff
@@ -1,3 +1,45 @@
...
+from django.dispatch import receiver
 
-# Create your models here.
+from .odoo import subscribe_user_to_news, unsubscribe_user_from_news
+
+
+class UserProfile(models.Model):
+    user = models.OneToOneField(User, on_delete=models.CASCADE, related_name="profile")
+    subscribe_to_news = models.BooleanField(
+        default=True, help_text="User agrees to receive Nextcloud news and updates."
```
edward-ly, line 12:
```suggestion
        default=True, help_text="User has opted in to receive Nextcloud news and updates."
```
and here

edward-ly (approved):

PR #1558, opened 2025-01-03
check if app is in hardcoded bundled list

tobiasKaminsky:


oleksandr-nc:
Thanks, Tobias, for taking care of this.

PR #1560, opened 2025-01-07
feat: re-add licenses by SPDX identifier, mark shorthand aliases as deprecated

### LINKED ISSUE #1274: More detailed licensing information
It would be helpful to include more detailed licensing information for all apps: E.g., AGPL-3.0-only or AGPL-3.0-or-later is a difference (I used [spdx](https://spdx.org/licenses/) license codes, here).

See https://github.com/NixOS/nixpkgs/commit/c317dcec0d36ba6c51f73bd23b5425a3f1d03141 for an example where this creates problems. In NixOS, the purpose is to package the apps so that Nextcloud can be declaratively configured and automatically deployed with particular apps.

edward-ly:
Resolves #1274.

Adds the SPDX identifiers for the currently supported licenses as valid enum values for the `<licence>` tag. Automatically converts `apache`, `mit`, and `mpl` to `Apache-2.0`, `MIT`, and `MPL-2.0`, respectively, when importing or creating a new app release. `agpl` and `gpl3` are left alone for now as they are implicitly assumed to map to `AGPL-3.0` and `GPL-3.0`, but those identifiers are also deprecated themselves.

Feel free to correct me if I'm wrong, but that was my interpretation based on the discussion so far.

Tasks for the app store admin (who I believe is @Altahrim?) after merging:

- [ ] rebuild app store documentation
- [ ] add the following rows to the `Licenses` table:

| id | name |
| -- | -- |
| AGPL-3.0-only | GNU Affero General Public License v3.0 only |
| AGPL-3.0-or-later | GNU Affero General Public License v3.0 or later |
| Apache-2.0 | Apache License 2.0 |
| GPL-3.0-only | GNU General Public License v3.0 only |
| GPL-3.0-or-later | GNU General Public License v3.0 or later |
| MIT | MIT License |
| MPL-2.0 | Mozilla Public License 2.0 |


nextcloudappstore/api/v1/release/info.xsd
```diff
@@ -369,6 +369,15 @@
 
     <xs:simpleType name="licence">
         <xs:restriction base="xs:string">
+            <xs:enumeration value="AGPL-3.0-only"/>
```
AndyScherzinger, line 372: best to add a comment that they are only used for `minVersion>=31` else you risk breaking instances with older XSD versions

nextcloudappstore/api/v1/tests/data/infoxmls/minimal.xml
```diff
@@ -5,7 +5,7 @@
     <summary>An RSS/Atom feed reader</summary>
     <description>An RSS/Atom feed reader</description>
     <version>8.8.2</version>
-    <licence>agpl</licence>
+    <licence>AGPL-3.0-or-later</licence>
```
AndyScherzinger, line 8: `minVersion>=31`

nextcloudappstore/api/v1/tests/data/infoxmls/fullimport.xml
```diff
@@ -16,7 +16,7 @@
     <version>8.8.2</version>
 
     <!-- only the AGPL license is possible -->
-    <licence>agpl</licence>
+    <licence>AGPL-3.0-or-later</licence>
```
AndyScherzinger, line 19: `minVersion>=31`

nextcloudappstore/api/v1/tests/data/infoxmls/app_api.xml
```diff
@@ -4,7 +4,7 @@
     <name>AppAPI Skeleton2</name>
     <description>Testing info.xml</description>
     <version>2.0.0</version>
-    <licence>AGPL</licence>
+    <licence>AGPL-3.0-or-later</licence>
```
AndyScherzinger, line 7: `minVersion>=31`

AndyScherzinger:
Besides my comments this looks good to me, yet this can only be merged _after_ the same PR has been merged on server v31.0.0, to ensure v31.0.0 (and later) won't break.

oleksandr-nc:
> Tasks for the app store admin (who I believe is @Altahrim?) after merging:

This will probably be done by me.

CI is expected to fail, it is broken due to invalid translations.

 I have informed @nickvergessen about this situation, for now I will temporarily remove the translations from the repository (but they will be returned by the bot again soon)

docs/developer.rst
```diff
@@ -335,7 +335,25 @@ version
...
+        * **MIT**
+        * **MPL-2.0**
+
+    * (deprecated) the following shorthand aliases are also used:
+
+        * **agpl** (AGPL-3.0)
+        * **apache** (Apache-2.0)
+        * **gpl3** (GPL-3.0)
+        * **mit** (MIT)
+        * **mpl** (MPL-2.0)
```
provokateurin, line 355: https://github.com/nextcloud/documentation/blob/master/developer_manual/app_development/info.rst this page must also be updated

edward-ly:
Andy's suggestions have now been applied. By the way, with the new license IDs, I've been thinking maybe we need to change the way the license is displayed in the server as well (e.g. [here](https://github.com/nextcloud/server/blob/c28fceb5d511e22030697e549c618699bce7c205/apps/settings/src/views/AppStoreSidebar.vue#L94)), but I'll leave it up to you whether anything needs to be done about it or not.

AndyScherzinger:
> I've been thinking maybe we need to change the way the license is displayed in the server as well (e.g. [here](https://github.com/nextcloud/server/blob/c28fceb5d511e22030697e549c618699bce7c205/apps/settings/src/views/AppStoreSidebar.vue#L94)), but I'll leave it up to you whether anything needs to be done about it or not

I had a similar thought if we should just drop the "-licensed" in the rendering

AndyScherzinger (approved):

nextcloudappstore/api/v1/tests/data/infoxmls/fullimport.xml
```diff
@@ -66,7 +66,7 @@
         <lib>iconv</lib>
 
         <!-- version numbers will be padded to three digits with 0 (min-version) and 2^64 (max-version) -->
-        <nextcloud min-version="9" max-version="10"/>
+        <nextcloud min-version="31" max-version="31"/>
```
oleksandr-nc, line 69: it is better not to change these values, or if you change them, then the tests should be changed as well to pass

edward-ly: Oops, I did forget to update the tests too, fixed.

oleksandr-nc (approved):
looks good to me

AndyScherzinger:
👍

provokateurin:
Btw can someone script PRs to update the license identifiers of all apps built by the company? If they were built after the OC split then we can safely assume that AGPL-3.0-or-later is meant. Alternatively if no direct change is possible due to ambiguity it would be nice to open an issue in each repo to warn about this.

PR #1561, opened 2025-01-08
fix: Show a note for bundled apps

nickvergessen:
grafik

Can't actually run this, screenshot is HTML adjusted from live page.

### DIFF
--- a/nextcloudappstore/core/models.py
+++ b/nextcloudappstore/core/models.py
@@ -320,14 +320,12 @@ class App(TranslatableModel):
             reverse=True,
         )
 
-    def is_outdated(self):
-        """Checks if an app has been released in last 3 recent platform versions
+    def is_server_bundled(self):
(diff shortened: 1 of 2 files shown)


nextcloudappstore/core/models.py
```diff
@@ -383,6 +381,18 @@ def is_outdated(self):
...
+
+    def is_outdated(self):
+        """Checks if an app has been released in last 3 recent platform versions
+
+        :return: True if not compatible, otherwise false
+        """
+
+        if self.is_integration:
+            return False
+        if self.is_server_bundled:
```
nickvergessen, line 395: Maybe
```suggestion
        if self.is_server_bundled():
```

as per pre-commit?
But not sure if that also works in the template

oleksandr-nc:
*I'll fix the python code + "pre-commit" if needed, a bit later in the evening*

oleksandr-nc (approved):

PR #1571, opened 2025-02-06
fix: send confirmation email for and verify updated email addresses

edward-ly:
This properly detects changes in a user's email address, sending a confirmation email that actually makes the change when the process is completed. Technically, `django-allauth` does not support "changing" an email address, only adding and removing them. As a workaround, the new address is added as the confirmation request is sent, and all other user addresses are deleted when the new address is confirmed.


oleksandr-nc (approved):
really great solution with "workaround", checked - seems to work 👍

PR #1572, opened 2025-02-10
fix(views): disable browser cache for user account pages

edward-ly:
This prevents user account information from being stored in the browser cache (which can still linger even after logout). This is especially useful when accessing the site on a shared computer.


oleksandr-nc (approved):
LGTM

PR #1650, opened 2025-09-22
fix: switch to outlined icons where possible

edward-ly:
... at least for the footer and the sorting options (with filled icons used to indicate selection for the latter). Most of the account menu icons do not have any outlined variants, and some of the app category icons need to be updated at a later date.

Sorting options:
Screenshot 2025-09-21 at 23-08-07 All apps - App Store - Nextcloud
Footer:
Screenshot 2025-09-21 at 23-02-10 All apps - App Store - Nextcloud


oleksandr-nc (approved):
looks good (tested on Safari)

PR #1671, opened 2025-12-04
chore: update app template

edward-ly:
Notable updates include migrating to TypeScript and Vue 3, removing webpack support, and updating the license tag.


oleksandr-nc (approved):
👍🏻

PR #1729, opened 2026-03-25
fix: add rate limiting to email update operation in account details page

edward-ly:
Currently, the default rate for this is approximately once every minute, which should be sufficient enough for our needs.


edward-ly:
Superseded by #1771.

PR #1750, opened 2026-05-05
chore(deps): npm audit fix

edward-ly:
Part of my recent efforts to catch up on dependency updates. The addition of the `punycode` package resolves a breaking change in webpack that previously included the package as a polyfill but was removed.


oleksandr-nc (approved):
Looks good!

PR #1773, opened 2026-05-20
docs(developer): add file size limit to app screenshots

edward-ly:
Related: https://github.com/nextcloud/usercontent.apps.nextcloud.com/pull/19

### DIFF
--- a/docs/developer.rst
+++ b/docs/developer.rst
@@ -438,6 +438,7 @@ repository
 screenshot
     * optional
     * must contain an HTTPS URL to an image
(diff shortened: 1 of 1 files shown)


miaulalala (commented):
10MB is a lot imho?

edward-ly:
> 10MB is a lot imho?

Hm, while I did want a bit of flexibility regarding allowing 4K screenshots, maybe I overestimated how much space that would take up. We can probably get away with setting the limit somewhere between 1-5 MB.

PR #1775, opened 2026-05-22
refactor(AccountForm): do not delete password errors manually

edward-ly:
This resolves an incompatibility with Python 3.14 where manipulating the Django forms `_errors` property directly is considered unsafe.

Assisted-by: Copilot:gpt-5.4


edward-ly:
Never mind, it turns out Django 4.2.x does not support Python versions higher than 3.12 ([ref](https://docs.djangoproject.com/en/6.0/faq/install/#what-python-version-can-i-use-with-django)). If we want to upgrade Python or Django later, we'll need to do both at the same time.

PR #1625, opened 2025-06-06
feat: add app authors to app details page

edward-ly:
Closes #891. Also displays any email addresses or websites specified with each author as appropriate.

Screenshot 2025-06-05 at 17-03-12 News - Apps - App Store - Nextcloud

PR #1796, opened 2026-06-17
feat: add GitHub download statistics for app developers

miaulalala:
## Summary

- Extends `GitHubClient` with `get_releases()` to fetch per-asset download counts from the GitHub API (and fixes a long-standing inverted auth header bug)
- Adds `get_download_counts()` which groups releases by `(owner, repo)` to minimise API calls; non-GitHub URLs return `null`
- New **API endpoint** `GET /api/v1/apps/{id}/downloads` — owner/co-maintainer only, returns JSON list with `version`, `is_nightly`, `download`, `download_count` per release
- New **web page** `/apps/{id}/downloads` with a table of the same data, linked from the app detail sidebar under a "Developer" section (visible only to owners/co-maintainers)

## Context

88.6% of the ~15k releases in the app store are hosted on GitHub using the standard `releases/download` URL pattern, so the GitHub API covers the vast majority of apps without any changes to how developers register releases. Releases on other hosts (GitLab, self-hosted, etc.) show `null`.

A `GITHUB_API_TOKEN` should be set in production settings to avoid the 60 req/hr unauthenticated rate limit.

## Test plan

- [ ] 18 new tests pass: `python manage.py test nextcloudappstore.core.tests.test_github nextcloudappstore.api.v1.tests.test_app_download_stats`
- [ ] Log in as app owner → app detail page shows "Developer / Download statistics" button
- [ ] Non-owners do not see the button and get 403 on the page/API
- [ ] Download stats page renders counts for GitHub-hosted releases; non-GitHub URLs show "— not available"


karlitschek:
yes please. I need this endpoint for some statistics too 🙏

nextcloudappstore/core/templates/app/detail.html
```diff
@@ -168,6 +148,16 @@ <h2>{% trans "Resources" %}</h2>
             </ul>
         </section>
         {% endif %}
+        {% if request.user == object.owner or request.user in object.co_maintainers.all %}
+        <section class="interact-section">
+            <h2>{% trans "Developer" %}</h2>
+            <a href="{% url 'app-downloads' object.id %}"
+               class="btn btn-default btn-light">
+                <span class="icon icon-chart-bar"></span>
```
edward-ly: In this repo, icons have to be committed into the repo and added to the CSS manually; otherwise, they won't be loaded into or appear in the app store at all. The process goes like this:

1. Download the SVG icon, in this case from here: https://pictogrammers.com/library/mdi/icon/poll/
2. Using Inkscape or your favorite vector graphics editor, re-scale the icon to 20px x 20px. Also, create two versions of the icon: `bar-chart.svg` where the icon is re-colored into the Nextcloud color (#007bbd), and `bar-chart-white.svg` where the icon is fully white (#fff) instead.
3. (optional) Copy and run the [image optimization script from the server](https://github.com/nextcloud/server/blob/master/build/image-optimization.sh) to compress the exported SVGs.
4. Add the necessary license information and CSS classes (including `icon-bar-chart`). You can use this commit as an example: https://github.com/nextcloud/appstore/pull/1625/changes/286f12d9f07049c22777f50ce13824791e757b82

edward-ly (changes requested):
Looks very nice! Just one small thing ...

miaulalala:
Thanks @edward-ly — done, following the mail icon in #1625:

- **`bar-chart.svg` / `bar-chart-white.svg`** in `img/icons/detail/`, matching the other MDI icons there (`bug-outline` and friends): `viewBox="0 0 24 24"` rendered at 20×20, `#007bbd` for the dark variant and `#fff` for the white one.
- **`icons.css`** in the three places that commit touched: the `--original-icon-*` `url()` definitions, the `--icon-*` mapping in `body`, and the `.icon-bar-chart` / `.icon-bar-chart-white` rules.
- **`REUSE.toml`**.
- Template updated from `icon-chart-bar` to `icon-bar-chart`.

Two things I did differently from the example, both worth your call:

**The REUSE entry is in the Austin Andrews block, not the Google LLC one.** `mail` is a Google Material Symbols icon so that was right for it, but this is a Material Design Icons glyph and MDI's `meta.json` records Austin Andrews as its author. Worth knowing: **`bug-outline`, also an MDI icon, currently sits in the Google LLC block** — that looks like a misfiling rather than a convention, but say the word and I'll move mine to match if it's deliberate.

**On the name** — you asked for `bar-chart` and that turns out to be the icon's own alias, which is a nice confirmation: MDI calls the glyph `poll` and lists `bar-chart`, `report`, `performance` and `analytics` as official aliases. So the CSS class is `icon-bar-chart`.

One note for completeness: `reuse lint` flags both new files exactly as it already flags `bug-outline`, so this matches the existing baseline rather than changing it either way.

edward-ly (approved):
🚀 

Quick note on the `bug-outline` icon: it is correctly filed in REUSE.toml, as evidenced by "Created by Google" noted [here](https://pictogrammers.com/library/mdi/icon/bug-outline/). I'm not sure what's going on with `reuse lint` when running locally, but as long as the CI passes, it should be fine, I think.

PR #1807, opened 2026-07-02
fix(user): allow GitHub-only accounts to edit profile without a password

miaulalala:
## Summary
- `AccountForm` unconditionally required a confirmation password when changing name/email/subscribe_to_news, but accounts created via GitHub OAuth have no usable password to confirm — so those edits always failed with "Password is required to change these fields."
- Fixed by checking `has_usable_password()` before requiring/validating the confirmation password. This is more general than special-casing the `github` provider (mirrors the approach `DeleteAccountForm` already uses for GitHub, but covers any social-only login).

## Test plan
- [x] Added `GitHubOnlyAccountEditTest` covering name and email changes for a GitHub-only (no usable password) account
- [x] `poetry run python manage.py test nextcloudappstore.user.tests --settings nextcloudappstore.settings.development` — all 15 tests pass

### DIFF
--- a/nextcloudappstore/user/forms.py
+++ b/nextcloudappstore/user/forms.py
@@ -108,8 +108,9 @@ class AccountForm(forms.ModelForm):
 
     def clean_passwd(self):
         value = self.cleaned_data.get("passwd")
-        # If password was entered, validate it
-        if value and not self.instance.check_password(value):
+        # If password was entered, validate it. Accounts without a usable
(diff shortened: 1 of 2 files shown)


edward-ly (approved):
While I don't think it's possible to test this in a local environment, the code makes sense. For a low impact change such as this, I think it's still fine to push this to production and try it out once it's deployed there.

And one small note, I added a `# nosec` in one of the user tests to suppress a false positive coming from Bandit.

PR #1840, opened 2026-07-16
feat: add is_enterprise_only attribute to apps

### LINKED ISSUE #1785: Implement "enterprise-only" app support
### How to use GitHub

* Please use the 👍 [reaction](https://blog.github.com/2016-03-10-add-reactions-to-pull-requests-issues-and-comments/) to show that you are interested into the same feature.
* Please don't comment if you have no relevant information to add. It's just extra noise for everyone subscribed to this issue.
* Subscribe to receive notifications on status change and new comments.

---

## Feature request

refs https://github.com/nextcloud/server/issues/60767

Each app would have an extra "enterprise-only" boolean attribute that can be toggled.
We could let the app owner toggle this attr when registering a new app or at any time later. Not sure the app owners need to change it at any time so: just in the app registration form would be enough.

Admin should be able to change this attribute at any time.

The apps list endpoint would filter out those enterprise-only apps by default.

When receiving a request for the app list, the appstore would support a new header, like `Subscription-Key`. On each request, the appstore would validate this key (if it's defined) and include the enterprise-only apps only if the key is valid.

TODO:

- [ ] Check how the portal validates/finds subscription keys
- [ ] Make a decision on the header

edward-ly:
Partially resolves #1785. In lieu of validating subscription keys, for now, a simple `include_enterprise` query parameter is added to the `/api/v1/apps.json` endpoint which controls whether or not enterprise-only apps are included in the list.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI


julien-nc (commented):
I'm not familiar with the appstore code so I had a look and: Lgtm.
Then I asked Qwen:
 
* urls.py:45 — appapi_apps.json uses apps_all_etag which now reads include_enterprise from the request. But AppApiAppsView doesn't filter by is_enterprise_only, so sending ?include_enterprise=true to that endpoint would compute an etag over a different dataset than what's actually returned. Either:
    - Give AppApiAppsView its own etag function that ignores the param, or
    - Also filter enterprise apps in AppApiAppsView if that makes sense for the app API.
* caching.py:42-57 — apps_etag and apps_last_modified still use App.objects.all(). The AppView (per-version) doesn't filter enterprise-only apps either. If the intent is to hide enterprise apps from the public app list, shouldn't the per-version endpoint also exclude them? This is the endpoint Nextcloud server actually calls to get compatible apps.
* The AppSerializer doesn't include is_enterprise_only, so API consumers can't tell which apps are enterprise-only. This might be intentional, but seems like an oversight — at minimum the Nextcloud server would need to know.
* views.py:198 — When an existing app owner re-registers (e.g. to update their certificate), is_enterprise_only gets overwritten from the request body. This means the owner can silently toggle the flag at any time via re-registration, which contradicts the issue's intent that only admins should be able to change it post-registration. Consider only setting it on creation
* The query param parsing is duplicated between AppsView._include_enterprise (views.py) and _include_enterprise (caching.py) with slightly different implementations. Extract to a shared utility.

Wdyt?

edward-ly:
All bullet points should now be addressed, thanks!

miaulalala (approved):
Clean implementation, good test coverage for the enterprise-only filtering (including etag/cache-busting edge cases). The existing-apps-unaffected behavior is correctly documented in the form help text. LGTM.

PR #1848, opened 2026-07-23
feat!(api): validate subscription key before retrieving enterprise-only apps

edward-ly:
Part 2 implementation of #1785. This replaces the `include_enterprise` toggle previously implemented in #1840.

## 🤖 AI (if applicable)

- [x] The content of this PR was partly or fully generated using AI

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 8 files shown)


edward-ly:
Finally tested on my local dev setup which actually works (aside from errors due to SSL cert validation which I had to disable, but I assume this only happens on local machines).

PR #1867, opened 2026-08-20
Feat: Add support for embedding videos in app store

### LINKED ISSUE #1507: Allow embedding videos on app store pages
Please allow app developers to include video presentations on the app store pages of their apps.

The text description and the screenshots that can be added currently are useful but in many cases they are not sufficient. People today are very visual and they would like to see a video presentation of the app, so that they can see the app in action and get an idea about its general look and its main features, before they decide if it fits their needs. The video presentation could be a youtube, vimeo, dailymotion, odysee, etc. video embeded on the page using an iframe, like [a video presentation](https://wordpress.org/plugins/duplicator/) on a WordPress plugin page.

Please consider accepting the `<video></video>` tag in the `appinfo/info.xml` files of apps, so that app authors can add short video presentations to their apps. If you think it's necessary you can restrict the source of the iframe video to some specific video platforms, restrict the duration of the video, etc.

Edit Daphne:
Embedding videos is a good idea but we do not accept embedding of Youtube, Vimeo, or other providers who add tracking. We would accept a PR that allows embedding a video in the app store from Peertube. This can be implemented probably directly into the app store and does not require changing the info.xml file.

lukasdotcom:
Adds support for embedding peertube videos from any instance.
You use it similar to screenshots:
```
<video>https://peertube.tv/w/28geJynHgsMqBhyPodg59Q</video>
```

image
Resolves: #1507 

## 🤖 AI (if applicable)

- [X] The content of this PR was partly or fully generated using AI


edward-ly:
Nice! Do you still have a copy of the test app so that I can try it out myself?

lukasdotcom:
> Nice! Do you still have a copy of the test app so that I can try it out myself?

The test app really just added a single video to the info.xml. Feel free to use it I already uploaded it to a git repo: https://github.com/lukasdotcom/junk/raw/1af52ee9a292b5d51767aed90af1928b709ffabd/testdummy.tar.gz

edward-ly (commented):
Works nicely on my end! Would it be possible to add an option in the user settings to enable autoplay on the first video when the page loads too? It should still be disabled by default, but if enabled, then maybe the video should be muted by default too (with another option to change this if desired).

lukasdotcom:
> Works nicely on my end! Would it be possible to add an option in the user settings to enable autoplay on the first video when the page loads too? It should still be disabled by default, but if enabled, then maybe the video should be muted by default too (with another option to change this if desired).

Added an autoplay feature that is off by default. The video has to be muted for autoplay anyway otherwise browsers might not autoplay.

Also added new video preferences section that currently just has autoplay
image

edward-ly (approved):
Awesome! 🚀

PR #1895, opened 2026-09-16
feat(admin): add administrator-settable release dates

miaulalala:
App store administrators are regularly asked to correct the date shown for an app, and there is currently no way to do it. On a release, every date is derived from when the upload reached the server. On an app, `last_release` is only ever written by the release importer — so an **integration**, which never uploads a release, shows the date it was registered and can appear years out of date while being actively maintained.

Both fields also feed the `apps.json` ETag, the per-app cache keys or the sort order, so making them editable directly would risk serving stale caches to every instance. This adds separate, purely presentational fields instead.

- adds an optional `release_date` to `AppRelease` and to `App`, nullable and blank
- exposes both in Django admin only, so app owners have no route to them (there is no writable serializer and no `PUT`/`PATCH` for releases)
- shows the release one on the releases page in place of the upload timestamp, and the app one as "Last updated" on the detail page, each when set
- leaves `created`, `last_modified` and `App.last_release` untouched, so caching, ordering and the API behave exactly as before
- tests, in separate commits, covering the fallback, the override, the unchanged ETag and the absence from the API serializers

Two things left deliberately for a follow-up: the exact semantics of the fields (a date the maintainer released upstream, versus a correction to a wrong or meaningless timestamp) and whether either should ever appear in the API.

Admin-only is intended rather than an oversight, and `<release_date>` in `info.xml` is an explicit non-goal — nothing reads a date out of the uploaded archive today, and if something did the field would be owner-controlled again.


edward-ly (approved):
🚀 Works nicely!

PR #1897, opened 2026-09-17
fix(api): return 405 when version kwarg is missing in AppView

### LINKED ISSUE #1896: GET on /api/v1/apps/<id> returns 500 instead of 405
`AppView` is bound to two routes with incompatible URL kwargs: `^platform/(?P<version>\d+\.\d+\.\d+)/apps\.json$` (`urls.py:43`, name `app`) and `^apps/(?P<pk>[a-z0-9_]+)/?$` (`urls.py:50`, name `app-delete`). Its `get()` starts with `version = self.kwargs["version"]` (`views.py:138-139`), so a `GET` on the second route raises `KeyError: 'version'` and returns **500 instead of 405 Method Not Allowed**.

Harmless to clients, but it fills the error log with tracebacks for what is just a wrong-method request. Seen in production on `GET /api/v1/apps/netbase`.

Ways to fix, cheapest first:

- Split the two routes onto separate view classes, so `AppView` only handles `DELETE` and the versioned listing lives in its own view. Removes the class of bug rather than the symptom.
- Keep one class but read the kwarg defensively, e.g. `version = self.kwargs.get("version")`, and return `405` when it is absent.
- Bind the versioned listing route to a dedicated `ListAPIView` and drop `get()` from `AppView` entirely.

Alibaba1234s:
### What
Return HTTP 405 Method Not Allowed when `version` kwarg is missing in `AppView.get()`.

## Why
Closes #1896. Currently `AppView` accesses `self.kwargs["version"]` directly. When bound routes without a version parameter (such as `apps/<pk>/`) receive a GET request, it raises an unhandled `KeyError` resulting in a 500 error instead of a clean 405 Method Not Allowed.

## How
- Used defensive retrieval via `self.kwargs.get("version")`.
- Returned `Response(status=405)` immediately if `version` is absent.

## Checklist
- [x] Follows CONTRIBUTING.md
- [x] Docs updated (if needed)


miaulalala:
Hey, thanks so much for taking a shot at fixing this issue! I tried to compile your changes locally but unfortunately it didn't build. Since there's also [another PR](https://github.com/nextcloud/appstore/pull/1899) superseding yours, I'm closing this one, but I'm nonetheless grateful and hope you may tackle another issue in the future. Cheers!

PR #1898, opened 2026-09-17
fix(core): widen Screenshot.url to 512 characters

miaulalala:
Migrations on production are blocked, so the store is stuck: `0037_convert_screenshot_urls_to_usercontent` aborts with `django.db.utils.DataError: value too long for type character varying(256)`, nothing after it runs, and `0038_video` never creates `core_video`. The same overflow also breaks release uploads at runtime, so apps with long screenshot URLs cannot publish.

## The arithmetic

0037 rewrites every `Screenshot.url` to `USERCONTENT_PROXY_URL + "/" + urlsafe_b64encode(source_url)`. The prefix `https://usercontent.apps.nextcloud.com/` is 39 characters, and base64 costs 4 bytes per 3, so a source URL of length *n* becomes `39 + 4 * ceil(n / 3)` characters. That crosses 256 at *n* = 163.

`info.xsd` caps a screenshot URL at 256 characters (`secure-url`), so the widest possible proxied value is `39 + 4 * ceil(256 / 3)` = 383. The column is widened to 512, which covers that with headroom. On PostgreSQL an increase of a varchar length is a catalogue-only change, so it does not rewrite the table.

The runtime path is the same transformation in `ScreenshotsImporter.get_proxy_url`, so widening the column fixes uploads too.

## Why 0037 is edited in place

The widening has to happen before the data migration in the graph, not after it. 0037 has never applied successfully in production - the `DataError` rolled its transaction back, so it is not recorded as applied - and the app store has a single deployment, so there is no other database holding a half-applied 0037.

## Tests

`nextcloudappstore/core/tests/test_screenshot_url_length.py`:

- the column fits the widest URL the schema allows, computed from `USERCONTENT_PROXY_URL` and the xsd limit rather than hardcoded, so it fails again if the column shrinks or the proxy host gets longer
- the 162/163 character boundary that used to overflow, kept as an explicit record of the regression
- a maximum-length proxied URL round-trips through the database
- `ScreenshotsImporter` stores a maximum-length source URL

Full suite: 218 tests, OK (2 skipped). `makemigrations --check --dry-run` reports no changes detected.


SystemKeeper (approved):
Don't know too much about the code, but looks sane

edward-ly (approved):
Thanks a lot! 🙏

PR #1899, opened 2026-09-17
fix(api): return 405 instead of 500 for unsupported methods on app routes

### LINKED ISSUE #1896: GET on /api/v1/apps/<id> returns 500 instead of 405

miaulalala:
Fixes #1896. `AppView` was bound to two routes with different URL kwargs, the versioned listing at `platform/<version>/apps.json` and app deletion at `apps/<id>`, and its `get()` read `self.kwargs["version"]` — so a `GET` on the delete route raised `KeyError` and returned 500, and a `DELETE` on the listing route failed the same way on the missing `pk`. Splitting the two means DRF answers any unsupported method with a 405 and a correct `Allow` header by itself, with no error handling to write.

Changes:

- Move the versioned listing into its own `PlatformAppsView`; `AppView` keeps only the destroy behaviour. `Allow` on `apps/<id>` goes from `GET, DELETE, HEAD, OPTIONS` to `DELETE, HEAD, OPTIONS`.
- Repoint the `app` route at the new view. URL names are unchanged, so nothing that reverses them moves.
- Add the direct assertions: `GET` on the delete route is 405, and its `Allow` header offers `DELETE` and not `GET`.
- Add a sweep that sends every method to every route in the v1 URLconf and fails on any 5xx, with a kwargs map that must be extended when a route is added.

Verified against production: `GET https://apps.nextcloud.com/api/v1/apps/twofactor_totp` currently returns 500, and something polls it. Nothing depends on that route serving `GET` — Nextcloud server only fetches `apps.json`, `categories.json` and `discover.json`, and the API docs list `DELETE` as the only method for `apps/{app-id}`.


nextcloudappstore/api/v1/views.py
```diff
@@ -126,12 +126,14 @@ class AppApiAppsView(EnterpriseFilterMixin, ListAPIView):
...
-        authentication.TokenAuthentication,
-        authentication.BasicAuthentication,
-    )
-    permission_classes = (UpdateDeletePermission,)
+class PlatformAppsView(GenericAPIView):
+    """Apps compatible with one Nextcloud version, served at platform/<version>/apps.json.
+
+    Kept apart from AppView on purpose: the two are bound to different routes with different URL
+    kwargs, and sharing one class meant a GET on the delete route reached a handler expecting a
+    "version" kwarg that route never supplies, raising KeyError and returning 500 instead of 405.
```
edward-ly:
```suggestion
```
This seems overly verbose to me. I'll add an `AGENTS.md` file later which may help prevent this.

edward-ly (approved):

PR #1901, opened 2026-09-17
fix(api): require authentication for write methods in UpdateDeletePermission

miaulalala:
`UpdateDeletePermission` implements only `has_object_permission`, and `BasePermission.has_permission` defaults to `True`, so the view-level gate is open and every check depends on the handler routing through `get_object()`. That is true of the generic views today, but it is implicit rather than stated, and a handler added later that does not fetch an object would run unauthenticated.

Stacked on #1899, which must merge first: on `master` this class still serves the public listing, and gating it there would break `platform/<version>/apps.json`.

Changes:

- Add `has_permission` to `UpdateDeletePermission`: write methods require an authenticated user, read methods stay open.
- Read methods deliberately stay open so a method the view does not implement still reaches dispatch and returns DRF's 405, rather than a 401 that hides it. Putting `IsAuthenticated` on the view instead does break that, because DRF runs `initial()` before handler lookup — tried, rejected, hence this shape.
- Add a test that an unauthenticated `DELETE` for a non-existent app is 401 rather than 404, which pins the ordering: the gate runs before the object is fetched.

`AppReleaseView` already pairs this class with `IsAuthenticated`; this makes the protection intrinsic to the permission rather than something each view has to remember.


nextcloudappstore/core/permissions.py
```diff
@@ -17,6 +17,21 @@ class UpdateDeletePermission(BasePermission):
...
+    def has_permission(self, request, view):
+        """Gate write methods at the view level, before any object is fetched.
+
+        has_object_permission alone is not enough on its own: BasePermission.has_permission
+        defaults to True, so every check depends on the handler routing through get_object().
+        That holds for the generic views today, but it is implicit, and a handler added later
+        that does not fetch an object would run unauthenticated.
+
+        Read methods stay open so that a method the view does not implement still reaches
+        dispatch and gets DRF's 405 rather than a 401 that hides it.
```
edward-ly:
```suggestion
```
This seems overly verbose to me. I'll add an `AGENTS.md` file later which may help prevent this.

